PR #31, opened 2015-01-21
Make the lift macro work in functions

tshort:
The existing `@lift` macro won't work in functions or other places where the types are not known at macro evaluation time. This pull request delays the `eval` until runtime to determine which variables are Signals. Use of eval is normally frowned on and is not fast, but for this application, it seems suitable. Setting up Signals isn't normally performance critical.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -15,15 +15,7 @@ function sub_val(ex::Expr, m::Module)
     elseif ex.head == :kw
         ex.args[2] = sub_val(ex.args[2], m)
     end
-    if ex.head == :call
-        try
-            eval(m, ex)
-        catch MethodError e
-            return ex
-        end
-    else
-        return ex
-    end
+    return ex
 end
 
 function extract_signals!(ex, m::Module, dict::Dict{Any, Symbol})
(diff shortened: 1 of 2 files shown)


shashi:
Looks good :) I'm going to make a release after fixing tests.

tshort:
Thanks, Shashi!

shashi:
@tshort I moved `test/macros.jl` to use FactCheck. I'm now getting weird variable not found errors with the macro. https://github.com/JuliaLang/Reactive.jl/pull/32/files#diff-3222be36181964907c68ce05ca638e79R8 from this line specifically. Looks like some escaping issue that I can't pin down. Could you give a shot at fixing this if you have a spare cycle or two?

shashi:
This breaks an example in Interact's demo:

```julia
In [21]: color = @lift RGB(r, g, b)
Out[21]:
4-element Array{Any,1}:
 RGB{T<:Union(FloatingPoint,FixedPoint)}                     
 Slider{Float64}([Input{Float64}] 0.5,"R",0.5,0.0:0.01:1.0)  
 Slider{Float64}([Input{Float64}] 0.5,"G",0.5,0.0:0.01:1.0)  
 Slider{Float64}([Input{Float64}] 0.79,"B",0.79,0.0:0.01:1.0)
```

It's probably my fault for removing `applicable(signal, x)` for `isa(SignalSource, x)`. Just making a note to fix this.

src/macros.jl
```diff
@@ -15,15 +15,7 @@ function sub_val(ex::Expr, m::Module)
...
-    if ex.head == :call
-        try
-            eval(m, ex)
-        catch MethodError e
-            return ex
-        end
-    else
-        return ex
-    end
+    return ex
```
shashi: @tshort This was there because if I had e.g. `sampleon(x, y)` I would want the whole expression to be a single input, without this, this will call `sampleon` with values of signals `x` and `y` which does not mean anything.

Was there some other reason you removed these lines?

Issue #25, opened 2014-09-23
Tutorial: ERROR: Lift not defined


kmsquire:
In http://julialang.org/Reactive.jl/index.html#do-you-even-lift:

```julia
julia> xsquared = lift(a -> a*a, Int, x)
4

julia> typeof(xsquared)
Lift{Int64} (constructor with 1 method)

julia> super(Lift{Int64})
ERROR: Lift not defined
```

`Lift` does not seem to be exported.

PR #32, opened 2015-02-19
Flatten. Fix #24, #25, `@lift` macro

### LINKED ISSUE #24: Feature Request: De/Attachable Subgraphs
Hi,
I've been speaking about this with Shashi, but I think so far I haven't had good examples, so the conclusion was mostly, that I'm using react wrong.
I did some more thinking, and I came to the conclusion, that we really need some mechanisms, to connect different react graphs, as you can't always construct the whole graph.
Example:
In an API I want to offer, to make a shader out of some string signal. But at shader creation, I don't want to force the user to decide, from which string signal he wants to create the shader. 

```Julia
function makeshader(defaultsource::String)
   sourcecode = ascii(defaultsource)
   dummyInput = Input(sourcecode)
   lift(dummyInput) do sourcecode
      ... Make new shader
      ... Delete old shader
      return shader
   end
   return dummyInput # return "socket" to which one can connect
end
# At initializsation, by some API, somewhere, where the user has no access
shader = makeshader(SomeShaderSource)

# At a different time point, user wants to interact with the shader, namely updating it
newsource = lift(x -> readall(open("filename")), every(1.0)) # this is just exemplary and a really wasteful way of doing this

# Now, this doesn't work in the current API:
lift(x-> push!(shader, x), newsource)
# So something like this is needed!?
connect(shader.dummyInput, newsource)
```
Another example:
I'm working on widgets for my plots.
The data for a plot in GLPlot looks a little like this:
```Julia
[
:attributename => Input(someValue)
:attributname => SomeLift
....
]
``` 
Now, at the part where I initialize this datastructure, I don't even remotely know, if the person wants to change the value via a my glwidgets, via the commandline with push!(), or if someone builds, for example, some tk wrapper.

Best,
Simon

shashi:
TODO

- [ ] Write tests for flatten


test/macro.jl
```diff
@@ -0,0 +1,43 @@
...
+
+        push!(a, 3)
+
+        @fact a.value^2 => b.value
+    end
+
+    # test use in a function
+    context("@lift inside a function") do
+        k = 3
+        f(a,b) = @lift a + b + 1 + k
```
shashi: @tshort This is still broken,

First, it says k is not defined. However, if I copy this to the REPL, the problem goes away. But then it screams a and b are not defined unless I have a and b in the outer scope, which does not feel correct...

I reverted some changes irrelevant to this feature from your PR, please take a look at the macros.jl in this branch. Was this case working when you tried??

PR #45, opened 2015-05-01
Include lift which works with non signals

SimonDanisch:
This is pretty convenient, if you don't want an anonymous function for every lift, that relies on a constant values.
Example:
```Julia
lift(+, some_array_signal, 1)
```
I'm not sure what the intended usage of `signal` is, but it would probably be better to overload it instead of using `makesignal`?

Best,
Simon

### DIFF
--- a/src/Reactive.jl
+++ b/src/Reactive.jl
@@ -337,6 +337,12 @@ lift(f::Callable, output_type::Type, inputs::SignalSource...; kwargs...) =
 lift(f::Callable, inputs::SignalSource...; kwargs...) =
     lift(f, map(signal, inputs)...; kwargs...)
 
+makesignal(s::Signal) = s
+makesignal(v) = Input(v)
+function lift(f::Callable, inputs...)
+    lift(f, map(makesignal, inputs)...)
+end
+
 # Uncomment in Julia >= 0.3 to enable cute infix operators.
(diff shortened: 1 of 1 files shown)


shashi:
I feel this method will make `lift` do more than one thing, which does not feel right.

How about adding an `always` signal primitive which holds a constant value (different from Input in that it cannot be pushed to). So this could be something like:

`lift(+, some_signal, always(1))` ?

`signal` is a function that can turn a `SignalSouce` to a `Signal`. This was added for convenience while working with widgets in Interact. Interact widgets are `SignalSource` subtypes.

SimonDanisch:
But would you then include the function like this:
```Julia
makesignal(x::Signal) = x
makesignal(x) = always(x)
lift(f, signals...) = lift(f, map(makesignal, signals)...)
```
Otherwise, it's just a little better than `lift(+, some_signal, Input(1))`, which was what I didn't like ;)
One reason I didn't mention yet is, that I also want to allow the user to pass either a signal or a constant value, without them having to explicitely write this out with `always` or `Input`.

shashi:
I'd keep this out of Reactive though, and use something like `makesignal` in the module that uses Reactive.

shashi:
I'll close this since you have const_lift ;) it seems like an impossible task now that we use the generic `map` function.

Issue #46, opened 2015-05-05
RFC: Signal type hell


shashi:
Passing in the right types to `lift`, `foldl` and `flatten` (and finally resorting to `Any`) in [these lines](https://github.com/shashi/Escher.jl/blob/3aaea4017d9662aa7dd8ceed75e0f934d9ee6077/examples/minesweeper.jl#L41-L48) caused me a *lot* of distress. This was mainly because a signal of type `Board{true}` will not update to `Board{false}`, and you are forced to annotate the type in each node in the subgraph.

This has got me wondering if there is *any* use to having the default type be the tightest possible type depending on the initial value in `lift`, `foldl` and `flatten`. It would be useful if one could statically warn of potential errors due to wrong types, Julia does not do this anyway. I am inclining towards using `Any` as default, and making tighter types as opt-in.

Also need consistency in the API for specifying types. I am thinking a `typ` keyword argument for `lift`, `foldl` and `flatten` can be consistent.

@stevengj @SimonDanisch what do you think of this?

shashi:
With the above commit, I was able to get away with: https://github.com/shashi/Escher.jl/blob/e47332d4db89940596c01a9f946b75b423518327/examples/minesweeper.jl#L41-L47 signal types of `lift`, `foldl` and `flatten` can now be specified with the `typ` kwarg. Suggestions to change this are welcome!

I had to mention the type inside the `foldl` function just once. I am really happy about this! Now to solve #47.

dcjones:
> I am inclining towards using Any as default, and making tighter types as opt-in.

I suppose you could also make tighter types default and opt-in for looser types, but I'm guessing the performance difference isn't a big deal so this sounds reasonable to me.

shashi:
@dcjones yeah, that's how it is right now (looser types are opt-in). The inconsistency in the API made it seem more complex and confusing. The only open question now is is `typ` a good name for the kwarg.

PR #48, opened 2015-05-05
Julia 0.4 compat.

dcjones:
Make things work on Julia 0.4.


dcjones:
Travis thinks I broke the tests on julia 0.3, but I'm not sure what to make of this:

```
ERROR: `lift` has no method matching lift(::Function, ::Type{Int64}, ::Input{Int64}, ::Lift{Int64})
```

shashi:
Oh that's me. I just broke the lift API. See #46 I'm going to rebase unless you have a suggestion for a better alternative to `typ`.

PR #56, opened 2015-06-14
Accomodate Timer API change

SimonDanisch:
Completely rewrote fpswhen. I'm not 100% sure, if it does exactly the same thing as the old fpswhen.
Also fixed type in timestamp and removed some anonymous functions.

### DIFF
--- a/src/timing.jl
+++ b/src/timing.jl
@@ -9,8 +9,7 @@ export every, fpswhen, fps, timestamp
 function every(delta::Float64)
     i = Input(time())
     update(timer) = push!(i, time())
-    t = Timer(update)
-    start_timer(t, delta, delta)
+    t = Timer(update, delta, delta)
     return lift(identity,  i) # prevent push!
 end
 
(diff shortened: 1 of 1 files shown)


src/timing.jl
```diff
@@ -9,9 +9,8 @@ export every, fpswhen, fps, timestamp
 function every(delta::Float64)
     i = Input(time())
     update(timer) = push!(i, time())
-    t = Timer(update)
-    start_timer(t, delta, delta)
-    return lift(x->x, i, typ=Float64) # prevent push!
+    t = Timer(update, delta, delta)
+    return lift(identity,  i) # prevent push!
```
SimonDanisch: > prevent push!
Do we need the same mechanism for fpswhen?
I'm not sure how this prevents "push!"

src/timing.jl
```diff
@@ -9,9 +9,8 @@ export every, fpswhen, fps, timestamp
 function every(delta::Float64)
     i = Input(time())
     update(timer) = push!(i, time())
-    t = Timer(update)
-    start_timer(t, delta, delta)
-    return lift(x->x, i, typ=Float64) # prevent push!
+    t = Timer(update, delta, delta)
```
sjkelly: I think this should be behind an `if VERSION >+ 0.4-<build>`.

src/timing.jl
```diff
@@ -22,31 +21,20 @@ end
 #     freq: the maximum frequency at which fpswhen should update
 # Returns:
 #     an signal of Float64 time deltas
+function gate(wason_timer::Tuple{Bool, Timer}, ison::Bool, s::Input{Float64}, delta::Float64)
+    wason, timer = wason_timer
+    (!wason&&ison) && return (ison, Timer(x->push!(s, time()), 0, delta)) # start pushing again
```
sjkelly: Parens are not needed here

src/timing.jl
sjkelly: The build is 5306

kmsquire: https://github.com/JuliaLang/Compat.jl/pull/103 should handle this once it's merged.

src/timing.jl
SimonDanisch: true - I started doing this to make it clear which part is the condition and which is the statement to execute.
Some of my weakly enforced coding standards you could say...

src/timing.jl
kmsquire: If you add a dependency to Compat 0.4.6 in `REQUIRE`, this should work on v0.3 as well.

PR #58, opened 2015-06-29
change REQUIRE to julia 0.3

tkelman:
unless anyone says otherwise that they're testing with 0.2-pre.


shashi:
Thanks! I'm a moron. Although I will miss sneaking this into METADATA

tkelman:
cargo cult copy-paste, it happens - at least until we excise these from all packages that aren't abandoned

PR #64, opened 2015-08-27
Docs, fix tests on 0.4

timholy:
Small things I noticed while kicking the tires. I can back out the `Do you even lift?` change if you want to keep that (I had to google for the meme in order to get it, but that just shows you how out of touch I am).


shashi:
No, I think derived signals is much better :)

Issue #62, opened 2015-07-28
Allow multiple signals updating at once


Keno:
When using signals heavily, it quickly becomes very annoying that only one signal can update at a time. What is the reason for this? I understand that you have to detect the same signal updating recursively, but why can't multiple independent signals update simultaneously? @shashi

shashi:
The restriction is pretty much only there to protect from recursive updates causing an infinite loop. You're right though that allowing multiple signals to propagate simultaneously is not that big a deal although it might mess up non-commutative `foldl`s...

The _correct_ way to implement signal propagation is using channels: each signal node reads from a channel of a parent and writes to a multi-cast channel subscribed to by all its children. This would take care that ordering is proper, and each node's function can be run in a different process if need be. Do you think it's worth trying? I have no idea what the overhead is going to be like when something like this is run on a single process, as is the case in Reactive's most common use cases.

Keno:
I don't understand why channels are required. It just seems like a per-signal isupdating flag would be enough for an initial implementation.

shashi:
Hmm, isupdating as a boolean flag or as a Condition? I don't understand how. Could you briefly elaborate?

Keno:
Well, I probably don't know the architecture well enough, but I was thinking just have a set of currently updating signals and don't allow push on those.

timholy:
Here's an application where this would be nice to have: https://github.com/timholy/GtkUtilities.jl (which has a Reactive-like component that I'd prefer to switch over to being based on Reactive) has properties for setting the x- and y- zoom range on a plot. (The docs are a little out of date, the current properties are `:viewx` and `:viewy`.) When a user draws a rubber-band selection region, it would be nice to be able to `push!(:viewx, newvaluex)` without triggering a Canvas redraw until I've also done `push!(:viewy, newvaluey)`. Right now I suppose the best way to do this is just `nx.value, ny.value = newvaluex, newvaluey` and then manually running the actions associated with `nx` and `ny`.

shashi:
How about adding a `push!{N}(::NTuple{N}, ::NTuple{N})` method?

timholy:
Works for me.

shashi:
I thought about implementing this, and I think this necessitates the heap that #65 got rid of.

Consider:

```julia
x = Input(1)
y = Input(2)

c = map(-, b)
d = map(tuple, a, c)
```
Initially `value(c)` = -2, `value(d)` = (1, -2)

[...]

Issue #61, opened 2015-07-16
fps and fpswhen don't return Δt


NOTtheMessiah:
It seems to create a signal of negatively accumulated timestamps instead of a Δt values

    julia> using Reactive

    julia> lift((delta) -> println(delta), fps(0.5))
    1.437081000427725e9
    nothing

    julia> -1.4370810026088e9
    -2.8741620071283436e9
    -4.311243013650205e9
    -5.74832402217328e9
    -7.185405032698582e9

tested in Julia 0.3.10

NOTtheMessiah:
Reverting this change to fpswhen should fix it: https://github.com/JuliaLang/Reactive.jl/commit/fa2dc44f54f4daa61b6598dbd9f00c50bb3c02e3

Issue #35, opened 2015-03-16
Data timing / windowing / buffering function


sirinath:
It is helpful to have function which:

 - Emit last value every n seconds if no new values
 - If there is multiple updates with the gap between the updates which are less than a threshold ignore and send last value in the cluster of updates
 - etc.

See:
http://reactivex.io/
https://github.com/GPars/GPars
http://hopac.github.io/Hopac/Hopac.html
https://github.com/reactor/reactor

shashi:
For the first one you can:

`sampleon(every(n), signal)`

I have plans to add windowing and throttling a PR is welcome though

shashi:
There is now `throttle` which can be used to get the same effect as `throttle` and `window` in Rx as far as I can tell

If there are more neat things that are worth having, please open a new issue.

PR #65, opened 2015-08-29
RFC: A simpler implementation

### LINKED ISSUE #54: Feature Request: Stopping a signal
image

Is it simple to implement an optional stop signal for every lift?
Background: if I animate a large array, e.g. particles, it would be great if I can stop it at some point and clean up the resources. 
Similar thing for cameras. It is sometimes simpler to just stop one camera and create a new one, instead of trying to recycle a pool of cameras.

Best,
Simon

shashi:
This is a rewrite of Reactive to make it more robust

# How it works

- A `Node` contains the current value and a vector of Actions
```julia
type Node{T}
    value::T
    actions::Vector{Action}
    alive::Bool
end

immutable Action
    recipient::WeakRef
    f::Function
end
isrequired(a::Action) = a.recipient.value != nothing && a.recipient.value.alive
```
- `add_action!(f, node, recipient_node)` creates an Action with ~~weakref to~~ recipient_node and function f. `send_value!(node, val, timestep)` sets val as node's value and calls all the actions with the `recipient_node` and `timestep` as the two arguments to `f`. `timestep` is a counter that increases with every new message.
- `connect_*` functions (e.g. `connect_map`) take both input and output nodes and add appropriate actions to them (spot the difference with [this example in sicp](https://mitpress.mit.edu/sicp/full-text/sicp/book/node64.html))
- exported functions like `map`, `foldp`, etc create a new output node and connect them to input nodes using the respective `connect_*` function 
- `_messages` is a global channel. `push!` writes `(node, new_value, metadata)` to the channel.
-  `Reactive.run(n)` reads n messages off this channel and propagates the changes.
   It throws a `ReactiveException`  with data to debug the crash. a `ReactiveException` prints a comprehensive crash message, for example:
```
ERROR: Failed to push!
    1
to node
    Node{Int64}(1, nactions=1)
at timestep 1

Debug info:
    ["meta"=>"kwarg","passed"=>"to push!","comes back"=>"here"]

error message goes here blah blah
 in buggy_function at /home/shashi/.julia/v0.3/Reactive/test/example.jl:10
 in anonymous at /home/shashi/.julia/v0.3/Reactive/src/operators.jl:39
 in send_value! at /home/shashi/.julia/v0.3/Reactive/src/core.jl:66
 in run at /home/shashi/.julia/v0.3/Reactive/src/core.jl:121
 in run at /home/shashi/.julia/v0.3/Reactive/src/core.jl:111
 in include at ./boot.jl:245
 in include_from_node1 at loading.jl:128
 in process_options at ./client.jl:285
 in _start at ./client.jl:354
 in run at /home/shashi/.julia/v0.3/Reactive/src/core.jl:125
 in run at /home/shashi/.julia/v0.3/Reactive/src/core.jl:111
 in include at ./boot.jl:245
 in include_from_node1 at loading.jl:128
 in process_options at ./client.jl:285
 in _start at ./client.jl:354
while loading /home/shashi/.julia/v0.3/Reactive/test/example.jl, in expression starting on line 20
```
- no `@lift` macro - it is too complex as it stands

# API changes

This is a breaking change

- Signal and Input are both typealiases to Node.`Input{Type}(val)` is now `Input(Type, val)`
- `lift` is now `map`
- `foldl` is now `foldp` - for fold-over-past
- `merge` breaks the guarantee of keeping the update from the earliest
  argument if more than one arguments update in the same timestep - the
oldest node gets priority (the propagation is depth-first)
- the main process needs to run `Reactive.run()` and watch for exceptions. A `ReactiveException` is raised which captures exceptions, backtraces and more info. exceptions don't get lost into the ether
- `push!` takes optional `meta` argument which shows up in the `ReactiveException` object if an  exception occurs
- `close(node)` makes the node refuse further updates
- added `delay`, `debounce` and `boundcopy`

You can call `push!` inside a `consume` this simply queues up the update for the next timestep.

This fixes #54 #62 #61 #24 

# TODO

- [x]  #35 - windowing/throttle
- [x] ~~right now the automatic gc'ing of nodes does not work: make it work with at least with some explicit api (relevant code is [here](https://github.com/JuliaLang/Reactive.jl/blob/b3f72ef14f9baa2a668f94215b73aed42959c3aa/src/core.jl#L27-L38))~~

- [x] Move old docs and change them

cc @SimonDanisch @Keno @stevengj @aviks @amitmurthy


src/core.jl
```diff
@@ -0,0 +1,137 @@
...
+    function run(steps=typemax(Int))
+        runner_task = current_task()::Task
+        local waiting, node, value, debug_meta, iter = 1
+        try
+            while iter <= steps
+                timestep += 1
+                iter += 1
+
+                waiting = true
+                (node, value, debug_meta) = take!(_messages)
```
timholy: It seems you could address #62 by doing an `isready` loop and setting the `value` of all the nodes at once while queuing the actions in a `Set`.

src/core.jl
```diff
@@ -0,0 +1,137 @@
...
+    end
+    nothing
+end
+
+include("exception.jl")
+
+# remove messages from the channel and propagate them
+global runner_task
+global run
+let timestep = 0
```
timholy: `timestep` to me implies something running on a timer. `counter`?

src/core.jl
```diff
@@ -0,0 +1,137 @@
...
+    recipient::WeakRef
+    f::Function
+end
+isrequired(a::Action) = a.recipient.value != nothing && a.recipient.value.alive
+
+type Node{T}
+    value::T
+    actions::Vector{Action}
+    alive::Bool
+end
```
timholy: I wonder if one could justify a second type, a `DeltaNode{T}` that signals only if the newly `push!`ed value is different from the old one.

timholy: OTOH there's a lot of value in simplicity...

timholy:
I'm pretty new to Reactive and haven't yet used it in a serious project, but a big :+1: for this. I'm especially excited to see the removal of all of those heap operations and like the `lift`->`map` rename. I'm planning on becoming an adopter.

While you're renaming things, I'll confess that the `push!`, `value` pair seems a little strange to me: I guess I kind of expect that `push!` implies a growing container, but a `Node` just stores a single value. What about `set!` and `get`? Or `put!` and `get`/`take`? But the status quo is fine if it makes more sense to you. Whatever words you choose here, the same ones should apply to widgets in Interactive.jl.

src/core.jl
SimonDanisch: I wonder if we should do these kind of things via properties or type parameters (I'd like signal that only updates for different values as well)

timholy: Yeah, a field makes sense. It's probably not worth the extra compilation overhead to do it with a new type or type parameter.

SimonDanisch: agreed! It also makes arrays of nodes more difficult to handle.

shashi:
@timholy great to hear! :)

I agree `push!` was ill-conceived. It seems `put!` or `set!` would indeed be better names for this. Both good candidates.

src/core.jl
shashi: will this be a whole lot better than
```julia
x = Input(0)
y = droprepeats(x)
```
?

src/core.jl
shashi: Updating many inputs at once may change some semantics. E.g. the value of `foldp(x -> x+1, 0, map(tuple, signal1, signal2))` will not be the same for the same sequences of input to signal1 and signal2 depending on how fast the runner is able to clear the queue... It also affects `sampleon`, `filterwhen` in subtle ways as well (but maybe these two are not as important as foldp)

timholy:
Would be good to also have the names reflect the "imperative"/"declarative" distinction clearly. `set!`/`put!` seems like it should take effect immediately to me, kind of like the implementation of `send_value!` (which to me implies something more like a remote-ref operation...). Now that I look at it, `push!` is not such a bad name for what it does now; however, one slightly confusing thing is that you're effectively pushing onto `_messages` rather than `n`---but it's complicated because it will _eventually_ `set!` `n`. `set!_queued(n, x)` or `queue_set!(n, x)`??

Wow, naming is hard.

timholy:
Or maybe `@queue set!(n, x)`. I like that better than my other two suggestions.

src/core.jl
timholy: Didn't realize that. Assuming that's functionally the same, then by all means let's do it that way.

src/core.jl
timholy: You understand the complexities much better than me. I'll simply add a use-case to #62.

shashi:
why a `@queue` macro when the job can be done by a function..? I think `queue_set!` is equally good, and better than `send_value!` :+1:

timholy:
I was thinking that `@queue set!(n, x)` would just rewrite the code as `queue_set!(n, x)`, but somehow I thought the macro syntax (a little remiscent of `@async`) made it clearer that you are queuing a `set!` operation.

timholy:
(Almost like `queue(set!(n, x))`, which of course wouldn't work.)

shashi:
Ah, I see the similarity with `@async`, but I think `queue_set!` will do here. It's not part of the public API anyway. One of the changes here (I just added to the list) is `foldl` being renamed to `foldp` - I'm more concerned how that will go...

src/core.jl
shashi: I find it nicer to say "at timestep X" rather than "when the counter was X", But yeah, probably not worth one extra definition...

timholy:
To clarify, I'm essentially proposing that
```
send_value! -> set! # you'll have to put this inside the let block so timestep can be filled in
push! -> queue_set!
```
I suspect this is inconsistent with `queue_set!` not being part of the public API. I'd argue you want an imperative form at your fingertips, and not rely on polling for everything.

src/core.jl
timholy: I'm fine with whatever.

shashi:
Oh I confused what you meant. I thought you were suggesting `send_value!` be renamed to `queue_set!`, well it now makes more sense, and so does `@queue set!` - because `queue_set!` can be confused for setting something on a queue... I think `enqueue!` and `push!` are very similar, so `push!` would work here as well...

timholy:
I'd be happy with `push!`. As I stated above, really you're pushing onto `_messages` rather than the `n`, but conceptually it's basically the same thing. One other issue, though, with the choice of `push!`: currently `push!(::Input, value)` is currently imperative, and you're proposing to make it be asynchronous. I don't know how much it matters, but that does seem like a tricky deprecation path, especially for people who are using this in event driven code (i.e., GUI toolkit handles the event loop and issues callbacks).

That said, we know which public packages use Reactive, so it's possible to ping their authors about this.

shashi:
Yes, this change for example affects [this Escher example](https://github.com/shashi/Escher.jl/blob/master/examples/mc.jl) by @aviks. It intermittently `push!`-es in a tight `for` loop. I think to alleviate this in most cases (where only one such simulation is running), you can run `yieldto(Reactive.runner_process)` to wake it up to read the message. It may be worth adding an optional boolean argument to `push!` for this.

timholy:
Well, in that case I'd just call `set!`---that's the new imperative function.

What I'm advocating is a good deprecation strategy. Now `push!` is imperative, and we're discussing changing the imperative to `set!` and making `push!` be asynchronous---which means that old code will not immediately throw an error, but may behave differently in ways that are hard to debug. A better approach might be to emit a deprecation warning for `push!` (encouraging users to switch to `set!`) and choose a different name for the asynchronous function (like `enqueue!(n,x)` or `@queue set!(n,x)`.

shashi:
I agree that we should have deprecation. I've added warnings for renamed / deleted functions. I'm not sure how to deprecate `push!` if the name and signature is not going to change... One more case I'm not sure about is `Input{T}(val) -> Input(T, val)` (I think the latter is more Julian, so decided to make the change)

about calling the thing `set!` - I don't think that's what it precisely does. You would be simply yielding to the runner task, but that doesn't guarantee that the next message it reads is the one put via the current push! call. (e.g. it could be a message from an `fps` signal, or if the runner task is already blocked elsewhere, there could be other messages)

timholy:
If that's how it works, I agree. I may have been slow to realize that you don't intend to have any events happen imperatively, that everything goes through the queue. So you can probably scratch a lot of what I said above (sorry).

src/core.jl
```diff
@@ -0,0 +1,137 @@
...
+            end
+        catch err
+            bt = catch_backtrace()
+            throw(ReactiveException(waiting, node, value, timestep, debug_meta, CapturedException(err, bt)))
+        end
+    end
+end
+
+# Run everything queued up till the instant of calling this function
+run_till_now() = run(queue_size(_messages))
```
timholy: `run_queued`?

shashi: Agreed run_till_now is not very clear. But `run_queued` sounds a lot like `run` I think. I'm fine with either. This was mainly meant for testing.

timholy: `push_and_run!(n, x; meta=nothing)`? Is there a circumstance where you'd need "run everything" separate from `push!`?

src/core.jl
```diff
@@ -0,0 +1,138 @@
...
+isrequired(a::Action) = a.recipient.value != nothing && a.recipient.value.alive
+
+type Node{T}
+    value::T
+    parents::Tuple
+    actions::Vector{Action}
+    alive::Bool
+end
+Node(x, parents=()) = Node(x, parents, Action[], true)
+Node{T}(::Type{T}, x, parents=()) = Node{T}(x, parents, Action[], true)
```
sjkelly: Would it be possible to switch this to:
```
Base.call{T}(::Type{Node{T}}, x, parents=()) =  Node{T}(x,parents,Action[], true)
```
So this can be backwards compatible on at least 0.4? e.g. `Input{Int64}(1)` will still work.

timholy:
Out of curiosity, what's your current thinking here? I've been planning to rebase GtkUtilities on Reactive, but had originally thought to wait until this was merged. If you have reservations about this PR, then perhaps I'll either proceed with the current Reactive or find another approach.

shashi:
@timholy Very sorry for the delay here. I'd like to get this merged. Right now the thing blocking this is GLVisualize examples which freeze with this branch but work if `gc_enable(false)` is run at the beginning. This is because of some node getting gc'd when it goes out of scope - a new _feature_ in this branch. I haven't been able to track where this is happening yet, but will take a closer look this weekend I'm looking forward to fixing this. I've migrated Escher to this branch and things work well.

 cc @SimonDanisch

src/core.jl
shashi: That's a good idea, I wasn't sure how to do it.

timholy:
No apologies needed, I just wondered if you were reconsidering. I have projects of my own that have lingered; sometimes it's just getting around to finishing, sometimes there's a blocker, but sometimes I become less sure if it's a good approach.

sjkelly:
+1 to merging to master and fixing the issues with GLVisualize before tagging. Chances are we might find the bug as other packages update to the new Reactive.

SimonDanisch:
Sounds like a reasonable strategie! I'd be surprised if the bug is so bad, that it makes the whole of reactive NEXT unusable ;)

shashi:
@timholy I still believe this is a strict improvement. :)

@sjkelly Okay I will rebase and merge soon

I'm trying to track down the GLVisualize issue by saving the backtrace when a Node is constructed. see [this commit](https://github.com/JuliaLang/Reactive.jl/commit/5d81185a9f4fb5434ffa52d14a545a7e110d8e7f)'s  (next-debug branch). [This gist](https://gist.github.com/shashi/882b17eedd4eac0a1087) runs a simple GLVisualize script and prints out the backtraces for cases where the node got gc'd...

but, the backtraces seem to be completely unrelated.... http://codepad.org/fJnYub1T

SimonDanisch:
How weird... `MacroTools` must stem from `Mustache.jl`, which uses `Requires.jl`, which uses `MacroTools`...

SimonDanisch:
I do some shady stuff with reactive and mustache right here:
https://github.com/JuliaGL/GLAbstraction.jl/blob/sd/reactive_next/src/GLShader.jl#L92

SimonDanisch:
I see if I can debug a bit as well ;)

shashi:
Update: I removed the `WeakRef` logic from nodes. This means all old code will be much happier. I realized this does not preclude the GC from helping out cleaning up nodes that go out of scope... Entire subtrees will get removed when all its nodes go out of scope! This is good. For example, with `Reactive.debug_memory=true`:

```julia
julia> using Reactive

julia> map(+, Input(1), Input(3))
Node{Int64}(4, nactions=0)

julia> gc() # might have to run this many times, but will eventually work
Node got gc'd. Creation backtrace:
call at /home/shashi/.julia/v0.4/Reactive/src/core.jl:49 (repeats 2 times)

julia> gc()
Node got gc'd. Creation backtrace:
call at /home/shashi/.julia/v0.4/Reactive/src/core.jl:49 (repeats 2 times)

julia> gc()
Node got gc'd. Creation backtrace:
call at /home/shashi/.julia/v0.4/Reactive/src/core.jl:50
 in _map at /home/shashi/.julia/v0.4/Reactive/src/operators.jl:19
 in map at /home/shashi/.julia/v0.4/Reactive/src/operators.jl:25

```

This means that when you reload a page in Escher, for example, the previous signals will go away. (I'm pretty sure this was the behavior in the current master as well... I have just been too dumb to think of it)


*However*, for some reason I cannot figure out, `every` or `fps` nodes refuse to go away...  Cleaning up these nodes is the most important (you want an animation to stop if no one is looking at it)

```julia
julia> every(1)
julia> gc() # as many times, but nothing happens Reactive.nodes still holds a reference to it...
```

I figured this should be because the timer callback points to the output node, so I tried [to use WeakRef in the callbacks](https://github.com/JuliaLang/Reactive.jl/blob/81f7eef05a8c0fa725061b32e6089635997ed33a/src/time.jl#L20-L35) but that does not help. `every_connect` does seem to not hold any reference to the Node and does what you expect, as can be verified by running something like: `Reactive.every_connect(1,Int[])`

Is there some way of checking what references an object has? That would help quickly figure out the problem.

(test failure is due to delay caused by printing of errors, will get fixed once I disable debug_memory flag)

timholy:
Are you sure you can do without `WeakRef`? I'm still trying to wrap my head around Reactive, but my concern is the following (illustrated by the `TwoSharedNodes` example in #75): suppose you start out with nodes `A` and `B` and `merge` them into `C`. Now later you want to replace `B` with `B2`; so you `close(C)` and then say `C = merge(A, B2)`. The problem is, `A` still holds an `Action` whose `recipient` is the *old* `C`, and so I don't think that old `C` can be garbage-collected.

timholy:
> However, for some reason I cannot figure out, every or fps nodes refuse to go away... 

With the new implementation, the timer never fires:
```
julia> using Reactive
INFO: Recompiling stale cache file /home/tim/.julia/lib/v0.4/Reactive.ji for module Reactive.

julia> n = every(1)
Node{Float64}(1.447256309980414e9, nactions=0)

julia> Reactive._messages
Channel{Any}(sz_max:1024,sz_curr:0)

julia> Reactive._messages
Channel{Any}(sz_max:1024,sz_curr:0)

julia> Reactive.nodes
WeakKeyDict{Any,Any} with 0 entries
```
(The delay there was much longer than a second.)

shashi:
This works fine for me....
```julia
julia> using Reactive

julia> every(1)
Node{Float64}(1.447269625871374e9, nactions=0)

julia> x=map(println, ans)
1.44726963488805e9
Node{Void}(nothing, nactions=0)

julia> 1.447269635888282e9
1.447269636888132e9
1.447269637888482e9
1.447269638889957e9
1.447269639890409e9
1.447269640891841e9
1.447269641893283e9
1.447269642893697e9
1.447269643895142e9
1.44726964489559e9
```

timholy:
Ah, you are correct.

I wondered if it might be a bug elsewhere, but this works:
```jl
julia> cleanup = Ref(false)
Base.RefValue{Bool}(false)

julia> z = [1]
1-element Array{Int64,1}:
 1

julia> y = Timer(x->WeakRef(z), 0.1, 0.1)
Timer(Ptr{Void} @0x00000000032e22a0,Condition(Any[Task (waiting) @0x00007f362f3d6d60]),true)

julia> finalizer(z, x->(cleanup[] = true; close(y)))

julia> isopen(y)
true

julia> z = 0
0

julia> gc()

julia> isopen(y)
false
```
So it seems like something else must be holding a reference?

shashi:
We seem to have correct implementations of everything we need now.

Couple of questions to answer to get this merged:

1. Do we need bottom-up GC of the graph like it was before dfabab7 ? @SimonDanisch doesn't seem to like sprinkling `preserve`s very much. @timholy what do you think?
2. The names `Signal`, `Node` and `Input` all point to the same thing... This one is a trifling matter, but it is slightly important that we bikeshed this. We ideally should to keep just one name, the best candidate seems to be `Signal`. We should deprecate `Input` for `Signal` if that makes sense to everyone (I feel the removal of `Input` will help understanding). `Node` is a good name internally (they are like wires which are connected to become or hold a signal) `Signal` is a good name in the public API. My odd preference here is to keep the type's name as `Node` but export `Signal` (typealias of `Node`). [Here is the code](https://github.com/JuliaLang/Reactive.jl/commit/cd9e83a1e2269897c33c9eab5ce9e99135db01f0). I will reluctantly agree to rename `Node` to `Signal` internally if that's what people want.

SimonDanisch:
1.)
Can you explain a little more? What's possible and what isn't right now? I think I'd be fine with sprinkling preserves, as long as they follow very understandable rules! If at some point the general rule of thumb would be: "rather put a preserve around your signal to be on the safe side", we'd done something wrong ;)
2.)
I agree with your preferences... `Signal` makes most sense to the user, while `Node` is a good name internally!

shashi:
1. Right now if a signal `a` is used to create several other signals, then either they all stop updating when they all go out of scope or all of them sit around and continue to sit around and update unless you `close` them explicitly. What's possible is: if `a` is used to create several signals, so as to make a graph of nodes, then any node whose children have gone out of scope (or a leaf node which has gone out of scope) will automatically stop updating after a gc pause. This means you will require `preserve` in places where you are doing side-effects e.g. `preserve(map(println, a))` instead of just `map(println, a)` so that the `map` node will continue to exist as long as `a` does. I don't know if any other reactive programming systems do this... I couldn't find out much by googling.
2. I'm relieved one more person agrees! :)

timholy:
> We seem to have correct implementations of everything we need now.

Congratulations!

> Do we need bottom-up GC of the graph like we had before dfabab7

I am torn on this one. On balance, I suspect that it would be better that way.

But I do have a question: suppose (like `fps`) we return a node/signal:
```jl
a = foo(7)
```
but internally in the course of creating `a`, we also need to create a "derived" node/signal `b` that we don't return to the user. We could, of course, call `preserve(b)`, but then I think that will keep `a` from being gced?


> Node/Signal/Input

Yes, definitely good to reduce the number of redundant names. You know about `@deprecate_binding`? (Used for deprecating types or other names.)

I agree with the preference `Node` internally. Externally, I have a slight preference for keeping it `Node` (rather than `Signal`) just for reasons of consistency (to facilitate blurring the lines between "user" and "developer" :wink:). But I will be fine with whatever you choose.

shashi:
> but internally in the course of creating a, we also need to create a "derived" node/signal b that we don't return to the user. We could, of course, call preserve(b), but then I think that will keep a from being gced?

I have another implementation of preserve which will allow `b` and `a` to be gc-ed together.

> You know about @deprecate_binding? (Used for deprecating types or other names.)

I didn't know about it, will use it. Thanks!

`Node` seems to be too general / meaningless a name to export...

shashi:
Here's the [preserve we need](https://github.com/JuliaLang/Reactive.jl/blob/bf1845ff56b2aa8f172216969dcdd4eed9783429/src/core.jl#L56-L87).

timholy:
I agree that `Node` is too general to export. That's a good reason for exporting `Signal`.

SimonDanisch:
Awesome! Thanks for the great work :)

timholy:
Yes indeed!

PR #75, opened 2015-11-11
Docs: add walk-through examples on Canvas redraws

timholy:
Note: this PR is against the "next" branch.

FRP is new to me, so as I struggle to wrap my head around it I thought I'd create a "diary." If that diary is useful to others, great; otherwise, feel free to close this.


doc/imageview_warmups.md
```diff
@@ -0,0 +1,329 @@
...
+julia> Reactive.run_till_now()
+
+julia> value(n)
+1
+```
+
+You can cause these updates to be processed at regular intervals---if
+julia isn't busy doing something else---as follows:
+```jl
+eventloop = Timer(_->Reactive.run_till_now(), 0.001, 0.001)
```
shashi: In the latest commits, I put this in [Reactive.__init__](https://github.com/JuliaLang/Reactive.jl/commit/2acb72efd6ce69f93e7b959eae9d69b19a4c77a8#diff-ac236c9f23e3c15683602f303d89d8cdR179).

The reason is, if one is using multiple packages that run `Reactive.run`, then they might end up stepping on each other's run calls. I also made it so that `push!(...)` takes an optional third argument which is a callback that will be called in case of an error... By default this callback prints the error to `STDERR`.

I was thinking the reason for exposing `Reactive.run` is to allow packages to watch for crashes, but this seems like a worse idea than having the onerror callback.

doc/imageview_warmups.md
```diff
@@ -0,0 +1,329 @@
...
+
+end  # module
+
+using Reactive, TwoSharedNodes
+
+tmr = Timer(_->Reactive.run_till_now(), 0.001, 0.001)
+
+n2 = Node(22)
+c1 = Canvas(STDOUT, "canvas1", 1, n2)
+c2 = Canvas(STDOUT, "canvas2", 2, n2)
```
shashi: I guess, it would be nice to have a way to `link` and `unlink` nodes in Reactive? so that when linked, updating one of them also updates the other...

I'm trying to understand this better, would it be fair to say that the code after this line is a simulation of user interacting with the Canvas UI?

doc/imageview_warmups.md
timholy: > In the latest commits, I put this in Reactive.init.

I was going to suggest that (for the reason you cite), glad you've already thought of it.

doc/imageview_warmups.md
```diff
@@ -0,0 +1,329 @@
...
+end
+
+node{T}(::Type{T}, val) = Node(T, val)
+node{T}(::Type{T}, n::Node{T}) = n
+
+function relink!(c::Canvas)
+    isdefined(c, :update) && close(c.update)
+    combined = merge(c.state1, c.state2)
+    throttled = throttle(1/60, combined)
+    c.update = map(x->println(c.io, "$(c.name): state1=$(value(c.state1)), state2=$(value(c.state2))"), throttled)
```
shashi: I see this line is where you would like the previous printing to stop...?

This is a good example of where we can't do without `WeakRef`. I'll think about this more

doc/imageview_warmups.md
timholy: > I guess, it would be nice to have a way to link and unlink nodes in Reactive? so that when linked, updating one of them also updates the other...

If it's easy to do, it seems like it could be nice. I do this in [GtkUtilities](https://github.com/timholy/GtkUtilities.jl#managing-state). For example, when I have a movie I'm looking at with 4000 frames in it, the slider is nice for uses but the entry box is best if you really need to go to frame 1374 and see what weird thing happened there...

GtkUtilities' API overall is not as flexible as Reactive. However, it is good for these kinds of "shared state" problems.

> I'm trying to understand this better, would it be fair to say that the code after this line is a simulation of user interacting with the Canvas UI?

Mostly "yes." One of the things I like about Reactive is that it seems to simplify having two points of control; this could be either "UI" or programmatic. But the main intent is to simulate user-driven updates. So each setting of `state!` might correspond to a click or drag in the UI; the "disconnect" could be triggered, for example, by right-clicking and picking an item on a popup menu.

doc/imageview_warmups.md
timholy: I should also say that I may be using Reactive differently than you intended...there clearly are differences in this style of thinking, and I'm only just getting used to it.

But yes, to me this seemed like a case where the `WeakRef` is needed if you want the nodes to be gc()ed. The `close` is presumably sufficient to stop the updates from "rendering," though.

doc/imageview_warmups.md
timholy: Oh, and if you think this has any tutorial value at all, I can make changes as you change the API. (I won't bother if you think it won't be useful, though.)

shashi: We should definitely update this for the new changes and merge this. This will be very useful for library writers using Reactive.

shashi:
Reactive is meant to be orthogonal to other packages. But when you are making a GUI-like package such as ImageView, you will need to think about precisely these things. It will be nice to have this documentation.

timholy:
Glad it seems useful. I will definitely revise this and resubmit.

doc/imageview_warmups.md
```diff
@@ -0,0 +1,329 @@
...
+    io::S
+    state1::Node{Int}
+    state2::Node{Int}
+
+    function Canvas(io::S, s1::Integer, s2::Integer)
+        n1, n2 = Node(Int, s1), Node(Int, s2)
+        c = new(io, n1, n2)
+        combined = merge(n1, n2)
+        throttled = throttle(1/60, combined)
+        Reactive.preserve(map(x->println(c.io, "state1: $(value(c.state1)); state2: $(value(c.state2))"), throttled))
```
shashi: I find the use of the `value` slightly awkward.... But it's really OK. Here's an alternative

`show_canvas(c) = map(render, sampleon(throttle(1/60, merge(c.state1, state2)), Node(c)))`

(I find it better to take out the rendering from the type definition as a personal preference)

shashi: Oh wait, you'd still need `value` inside render. Never mind me.

timholy:
Indeed it's much simpler now.

musm:
bump

PR #77, opened 2015-11-14
Permit gc on fps nodes

timholy:
The tests don't pass, but hopefully you can fix that? This version seems to allow cleanup of fps-created nodes. As with #76, it's not really obvious to me why this works when the previous one didn't. But I at least find this version easier to read anyway, so hopefully this is a step forward.


shashi:
I'm waiting to see if this one passes tests.

timholy:
Unless you made changes somewhere, this won't pass tests without further bugfixes. Something is broken with both `fps` and `every`, but I don't have long enough experience to understand how it's supposed to work.

shashi:
Yeah I was trying to fix that and failing to reproduce similar failures in the REPL. Will take a look again.

PR #79, opened 2015-11-21
undo call{T}(::Type{Node{T}}, x) deprecation

sjkelly:
This undoes the deprecation introduced in #65 for `call{T}(::Type{Node{T}}, x)`. AFAIK it wasn't an issue with syntax as much as the difficulty of finding the right `Base.call` (arggghh type invariance!).

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -48,8 +48,18 @@ immutable Action
 end
 isrequired(a::Action) = a.recipient.value != nothing && a.recipient.value.alive
 
-Signal{T}(x::T, parents=()) = Signal{T}(x, parents, Action[], true, Dict{Signal, Int}())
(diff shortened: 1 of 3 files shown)


shashi:
Is there a way to keep `Signal(T, value)` working?

I think this is consistent with the Julia convention, e.g. `Array(T, 10)` etc.

sjkelly:
Yes, I can do that. Since 0.4 `Array{T}(10)` has been supported with call, so it is helpful to have both I guess.

sjkelly:
I lost some motivation rebasing for the `Signal` rename, so I just removed the deprecation and added constructor tests. This makes the patch much smaller.

sjkelly:
Bump.

PR #91, opened 2016-04-08
better stop and start of event loop

SimonDanisch:
I changed the while loop to a for loop, since that seems to represent better what it is... Hope that's okay ;)

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -178,12 +178,8 @@ _push!(::Void, x, onerror=print_error) = nothing
 global run
 let timestep = 0
     function run(steps=typemax(Int))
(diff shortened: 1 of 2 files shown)


SimonDanisch:
Is throttle broken on 0.5, or is this just some fluctuation?
Removing `sleep` in `test/time.jl:93` seems to make the tests pass...

shashi:
Weird how travis fails for all versions though...

SimonDanisch:
Ah yeah on OSX... I just can't reproduce it locally ;)

shashi:
looks like the increased compile time is causing the hand-wavy timing tests to fail.

PR #96, opened 2016-05-15
Add explicit dependency on Compat

tkelman:
Pkg.test was only passing because FactCheck is in test/REQUIRE
and FactCheck happens to depend on Compat


shashi:
Oops, didn't see it, made my own commit. Thanks anyway!

PR #97, opened 2016-05-24
Add option to specify type of throttle Signal

timholy:
As illustrated in the test, I discovered that `throttle` on a `merge`ed signal can run into trouble if the different signals have different types.

BTW, the CI failure is something that also happens for me locally on occasion. Seems like there might be some kind of race condition in that test?

### DIFF
--- a/src/deprecation.jl
+++ b/src/deprecation.jl
@@ -10,4 +10,6 @@ export lift, consume, foldl, keepwhen, keepif, dropif, dropwhen
 @deprecate keepif filter
 @deprecate dropif(f, default, signal) filter(x -> !f(x), default, signal)
 @deprecate dropwhen(predicate, x, signal) filterwhen(map(!, predicate), x, signal)
-@deprecate call{T}(::Type{Signal{T}}, x) Signal(T, x)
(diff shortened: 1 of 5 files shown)


timholy:
Hmm, several other failures. There seems to be some kind of flakiness that manifests on Travis?

shashi:
A signal in the test was going out of scope and getting gc'd. Should be fixed now. Waiting for CI.

timholy:
Awesome!

I suspect the other errors (which all seem to be errors of timing) are due to the fact that these machines are presumably shared; `sleep(0.1)`, while usually good to within 10% or so, was occasionally bigger than 0.13.

timholy:
Looks like it happened again.

shashi:
I don't know if naming the signal prevents gc because it is not used later in that scope. I'm trying `preserve` instead, which shouldn't have this problem.

timholy:
Seems like just holding a reference to it should do the trick, but I do see a lot of green here.

Since you know much more about this package than me, I'll let you decide about merging. Thanks for taking the time to chip in!

PR #98, opened 2016-05-27
Update index.md

hgeorgako:
Two changes here: One on line 196 where the value should be updated to 6, and the other for the droprepeats() command which was mentioned but not shown in the code snippet (please verify this is correct).


shashi:
thanks!

PR #107, opened 2016-09-20
fix stop

SimonDanisch:
Does this look reasonable?


shashi:
Seems OK. Any idea why tests fail on nightly?

SimonDanisch:
Ugh, throttle again :D No idea, yet... Let me see if I can reproduce it locally and if master fails as well...

SimonDanisch:
Hm can't reproduce on ubuntu on 0.6 and 0.5 :(

Issue #110, opened 2016-10-06
Use appveyor


shashi:

shashi:
@tkelman, I can't figure out how to flip the switch on Appveyor, could you help?

tkelman:
What have you tried.

shashi:
There was no "New project" button when I logged in through JuliaLang. I could do it with my own account.

PR #111, opened 2016-10-06
activate appveyor

SimonDanisch:
fix for #110


SimonDanisch:
I've added Reactive to appveyor, but it seems like it doesn't pick it up yet. Maybe this needs to be on master first?

shashi:
Thanks! Let's find out :)

Issue #15, opened 2014-08-04
React Benchmark


SimonDanisch:
Hi,
I made a little benchmark, to see how big the performance gap is between React and "unrolled" code.
Here's the code
https://gist.github.com/SimonDanisch/fe5eaa964a5a1c197f01
Results:
unrolled: 6.8611264019997655e-6
React: 6.653197434700169e-5
Which isn't too bad, considering what React does.
I'm wondering, if my benchmark code is a good comparison though.
Also, I want to start a discussion on how to make React faster ;)
Maybe we can start to emit customized code, instead of putting the lifts in data structures?

PR #113, opened 2016-11-16
Use an action queue/set to ensure actions are performed in the correct order and aren't repeated

### LINKED ISSUE #101: Signal not updating correctly?
Consider the following code

```
using Reactive

x1 = Signal(1)
x2 = Signal(10)
y1 = map(x1) do x println("computing y1"); x end
y2 = map(x2) do x println("computing y2"); x end
y12 = map(x1, x2) do args... println("computing y12"); +(args...) end
z = map(y1, y2, y12) do args... println("computing z"); +(args...) end
```

So z is the sum of ys and the ys depend in some way on the xs. Now lets update this:

```
push!(x1, 3)
sleep(1)
@show value(x1)
@show value(x2)
@show value(y1)
@show value(y2)
@show value(y12)
@show value(z);
```

which gives me

```
computing y1
computing z
computing y12
value(x1) = 3
value(x2) = 10
value(y1) = 3
value(y2) = 10
value(y12) = 13
value(z) = 24
```

All values look as I expected except for z. It was defined as sum of ys, so it should be 26. z should have been updated a second time after y12 changed right? Or do I misunderstand how Reactive.jl is supposed to work?

JobJob:
Instead of performing actions attached to a node as soon as a new value is sent to it, add the actions to an ordered set `action_queue` for the current `push!` message, keyed (de-duplicated) using the recipient node. This ensures that actions that would previously have been performed twice, or performed once before all the inputs to the action had updated (and ignored later after they had all updated due to the `timestep` mechanism), will be performed once and at the correct time, i.e. after all their inputs have updated for this `push!` message.

Essentially causes the Signal graph to be traversed breadth first, rather than the previous depth first.

Fixes #101

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -1,6 +1,8 @@
 import Base: push!, eltype, close
 export Signal, push!, value, preserve, unpreserve, close
 
+using DataStructures
+
 ##### Signal #####
 
 const debug_memory = false # Set this to true to debug gc of nodes
@@ -46,6 +48,9 @@ immutable Action
     recipient::WeakRef
     f::Function
 end
+
+const action_queue = OrderedDict{Signal, Action}()
+
 isrequired(a::Action) = a.recipient.value != nothing && a.recipient.value.alive
 
 Signal{T}(x::T, parents=()) = Signal{T}(x, parents, Action[], true, Dict{Signal, Int}())
@@ -127,7 +132,8 @@ function send_value!(node::Signal, x, timestep)
     # Set the value and do actions
     node.value = x
     for action in node.actions
-        do_action(action, timestep)
+        action.recipient.value != nothing && #nothing means downstream node has been gc'd
+            (action_queue[action.recipient.value] = action)
     end
 end
 send_value!(wr::WeakRef, x, timestep) = wr.value != nothing && send_value!(wr.value, x, timestep)
@@ -201,7 +207,11 @@ let timestep::Int = 0, stop_runner::Bool = false
         post_empty() # make sure it's not waiting on an empty message list
     end
 
-
+    Base.shift!(od::OrderedDict) = begin
+        firstkey = od |> keys |> first
+        res = od[firstkey]; delete!(od, firstkey)
+        firstkey=>res
+    end
 
     """
     Processes `n` messages from the Reactive event queue.
@@ -213,6 +223,10 @@ let timestep::Int = 0, stop_runner::Bool = false
                 isa(message, EmptyMessage) && continue # ignore emtpy messages
                 try
                     send_value!(message.node, message.value, timestep)
+                    while length(action_queue) > 0
+                        (node, action) = shift!(action_queue)
+                        do_action(action, timestep)
+                    end
                 catch err
                     if isa(err, InterruptException)
                         println("Reactive event loop was inturrupted.")
--- a/src/time.jl
+++ b/src/time.jl
@@ -8,7 +8,7 @@ Throttle a signal to update at most once every dt seconds. By default, the throt
 This behavior can be changed by the `f`, `init` and `reinit` arguments. The `init` and `f` functions are similar to `init` and `f` in `foldp`. `reinit` is called when a new throttle time window opens to reinitialize the initial value for accumulation, it gets one argument, the previous accumulated value.
 
 For example
-    y = throttle(0.2, x, push!, Int[], _->Int[])
+    `y = throttle(0.2, x, push!, Int[], _->Int[])`
 will create vectors of updates to the integer signal `x` which occur within 0.2 second time windows.
 
 """
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
 julia 0.4
+DataStructures
--- a/test/basics.jl
+++ b/test/basics.jl
@@ -61,7 +61,7 @@ facts("Basic checks") do
         push!(a, number())
         step()
         # precedence to b over a -- a is older.
-        @fact value(e) --> value(a)
+        @fact value(e) --> value(b)
 
         c = map(_->_, a) # Make a younger than b
         f = merge(d, c, b)
@@ -73,6 +73,7 @@ facts("Basic checks") do
     context("foldp") do
 
         ## foldl over time
+        gc()
         push!(a, 0)
         step()
         f = foldp(+, 0, a)
--- a/test/call_count.jl
+++ b/test/call_count.jl
@@ -35,3 +35,40 @@ facts("Call counting") do
         @fact cf.value --> 2i
     end
 end
+
+facts("multi-path graphs") do
+    a = Signal(0)
+    b = Signal(0)
+
+    c = map(+, a, b)
+    d = merge(a, b)
+    e = map(+, a, map(x->2x, a)) # Both depend on a
+    f = map(+, a, b, c, e)
+
+    for (av,bv) in [(1,2),(1,3),(7,7)]
+        push!(a, av)
+        push!(b, bv-1)
+        Reactive.run_till_now()
+        push!(b, bv)
+        Reactive.run_till_now()
+        @fact value(c) --> av + bv
+        @fact value(e) --> 3av
+        @fact value(d) --> bv
+        @fact value(f) --> 5av + 2bv
+    end
+
+    xv = 2
+    x = Signal(xv)
+    y = map(identity, x)
+    x2 = map(x->2x, x)
+    z = map(+, y, x2)
+    Reactive.run_till_now()
+    @fact value(y) --> xv
+    @fact value(x2) --> 2xv
+    @fact value(z) --> 3xv
+
+    xv2 = xv + 1
+    push!(x, xv2)
+    Reactive.run_till_now()
+    @fact value(z) --> 3xv2
+end


src/time.jl
```diff
@@ -8,7 +8,7 @@ Throttle a signal to update at most once every dt seconds. By default, the throt
 This behavior can be changed by the `f`, `init` and `reinit` arguments. The `init` and `f` functions are similar to `init` and `f` in `foldp`. `reinit` is called when a new throttle time window opens to reinitialize the initial value for accumulation, it gets one argument, the previous accumulated value.
 
 For example
-    y = throttle(0.2, x, push!, Int[], _->Int[])
+    y = throttle(0.2, x, push!, Int[], \_->Int[])
```
JobJob: This was causing my editor to get confused because it interprets it as markdown.

shashi: :+1:

src/core.jl
```diff
@@ -201,7 +206,11 @@ let timestep::Int = 0, stop_runner::Bool = false
         post_empty() # make sure it's not waiting on an empty message list
     end
 
-
+    Base.shift!(od::OrderedDict) = begin
+        firstkey = od |> keys |> first
+        res = od[firstkey]; delete!(od, firstkey)
+        firstkey=>res
+    end
```
shashi: should probably upstream this, but fine here for now

src/core.jl
```diff
@@ -1,6 +1,8 @@
 import Base: push!, eltype, close
 export Signal, push!, value, preserve, unpreserve, close
 
+using DataStructures
```
shashi: needs to be added in REQUIRE

src/core.jl
```diff
@@ -127,7 +132,7 @@ function send_value!(node::Signal, x, timestep)
     # Set the value and do actions
     node.value = x
     for action in node.actions
-        do_action(action, timestep)
+        action_queue[action.recipient.value] = action
```
shashi: what happens if the value is `nothing`?

SimonDanisch:
Is there any performance tax? I wanted to benchmark this when ready.
I'd hate to have any performance tax to be honest. As it is, Reactive is already too slow for my taste. 
But to be honest, I should also refactor my code to use less lifting and less `push!`.

src/core.jl
JobJob: 👍

src/core.jl
JobJob: `action_queue` keys must be `Signal`s so: `MethodError: Cannot `convert` an object of type Void to an object of type Reactive.Signal{T}`

But I don't think it will happen, unless someone is doing something with non-exported functions/variables. Actions only get created and added to `node.actions` in `add_action!` (which is a nice feature of the existing design btw - made this change easy).

src/core.jl
JobJob: 👍

src/time.jl
JobJob: Actually, I can just wrap the function call in backticks. See updates.

JobJob:
Yeah, this will likely have a (hopefully small) performance hit, though maybe some will be regained through not calling some Action functions more than once, e.g. the examples in #101. Without it though, there will be subtle bugs.

When I was thinking about this, I realised that the sequence of Actions/functions that gets called each `push!` can largely be pre-determined from the Signal graph. I implemented a version of this that stores an action_queue for each node that gets pushed to. It was going well until I realised that with `filter` and `flatten` things get a little complicated. Anyway, I persisted and got it passing all the tests, but the code in parts is a bit ugly and complicated - [branch here] (https://github.com/JobJob/Reactive.jl/tree/action_queues_redesign), small design doc [here](https://github.com/JobJob/Reactive.jl/blob/action_queues_redesign/doc/action%20queue%20design.md). It needs a bit more work - I'm not 100% confident in its correctness, and the design of some of the operators (esp. `filter` and `flatten`) is not very clean, though I would be curious to see if it's faster, it probably should be, or could be made to be.

In general, I think the problem can be thought of as finding an optimal execution plan for a computation graph, so maybe we can leverage other work to do it faster.

So anyway, @SimonDanisch, I'm very interested in a performance benchmark.

src/core.jl
JobJob: Having said that, it seems to be happening on these builds: https://ci.appveyor.com/project/shashi/reactive-jl/build/1.0.4/job/o44rh18pio50ok8r
https://ci.appveyor.com/project/shashi/reactive-jl/build/1.0.4/job/63i6ry0w3fgbxm94

😃

Not sure how...

JobJob: Oh of course, it was getting gc'd. Fixed now, I believe, with an explicit test for the `recipient.value != nothing`

JobJob:
Anyway, I think this should probably be merged, since without it you can get weird bugs.

I think we can address performance separately, since it's not clear where the major bottlenecks are, nor whether this reduces performance significantly.

shashi:
We can address performance now.

JobJob:
We need a benchmark. A realistic one would be good, looking at you @SimonDanisch. I've only used Reactive for simple UI stuff with just a few signals thus far.

SimonDanisch:
let me check it out with GLVisualize today ;)

SimonDanisch:
uh, results look pretty bad... In my simple benchmark it was a 7x slow down before this PR.
Now it's 70x slow down.
Most time seems to be spent in: https://github.com/JuliaGizmos/Reactive.jl/blob/master/src/core.jl#L210
But I guess one could optimize the implementation a bit?
I'm on a pretty beefy machine right now and I don't have any GLVisualize benchmark pushing Reactive to the extreme so I can't really judge the impact on that right now.

SimonDanisch:
The famous, very old benchmark:
https://gist.github.com/SimonDanisch/fe5eaa964a5a1c197f01#file-reactbench-jl
(#15)

JobJob:
Ok cool, nice one. 

Yes it can definitely be optimised. 

For starters you don't even actually need an ordered dict, you could use an array of some fixed size (grow it if necessary), and keep track of how many actions you are adding each push!, and just run through actions 1:num_actions_this_push

JobJob:
Now there's a benchmark, we can all have fun seeing if we can shave some time off it 😏

shashi:
A simpler benchmark

```julia
x = Signal(0)
y = map(-,x)
@benchmark (push!(x, 2); Reactive.run_till_now())
```
pretty much 90% of the time is spent in `shift!`. I think it might have to do with it not being able to infer the return type.

shashi:
Ahh, just using a Queue instead of OrderedDict removes all the stupid allocations.

@SimonDanisch can you check now? The overhead to this is about 60ns amortized on my machine.

JobJob:
Actually, there was a reason I used an OrderedDict. It was to avoid an action being added to the queue more than once per `push!`. The reason the Queue change works is:

1. the queue ensures the actions are performed in the order they are added to each node, which is the correct (topological) order, whereas before the actions were done depth first.
1. the timestep check ensures actions aren't performed more than once, even though multiple actions are added to the queue for each node (e.g. one for each input to a map node).

JobJob:
So I woke up this morning in a sweat because I realised that this change, with the OrderedDict, and Shashi's Queue modification, aren't actually running the nodes in topological order! 😄 . They're effectively using a breadth-first order, which was better than the previous depth first, but still has bugs in some cases. E.g.:
```
xv = 2
x = Signal(xv)
x11 = map(identity, x)
x12 = map(identity, x)
x2 = map(+, x11, x12) # == 2x
x3 = map(+, x, x2) # == 3x
```
This is 6 at first (which is correct) but after a `push!(x, 3)` it updates to a 7 not a 9. Nodes are being run in order of their minimum depth from the node being pushed to, not their maximum depth.

Nodes are always created in topological order (since you can't create a node that has an input that isn't created yet), and the branch I mentioned above ([code here](https://github.com/JobJob/Reactive.jl/tree/action_queues_redesign)) adds actions to an action queue/list (one queue for each top level node) in the order the nodes are created and works for the above case. 

As I mentioned above, the code there is more of a proof of concept, it needs some cleaning up, but I think it should be able to be made slightly faster than the current Reactive:master implementation, since the action queue is only modified when nodes are added, not during `push!`. When a node is `push!`ed to it just runs through the node's action queue in order.

A potential disadvantage of the current implementation in that branch is that you can't `push!` to non top-level nodes, it could be added though, but not sure if this is desired.

Anyway, I probably won't have time to clean it up until maybe next week, but since people seem to be looking at optimisations right now, I thought I should mention it.

JobJob:
Sorry, haven't had time to clean up my branch - not sure when I will unfortunately - I was hoping to come up with a way to simplify the `filter` and `flatten` implementations, but nothing super simple has struck me, and haven't had too much time to work more on it.

Furthermore the issue is worse than I originally thought. For instance, this simple case is badly behaved after this PR 😮 
```
xv = 2
x = Signal(xv)
x2 = map(x->2x, x) #2x
x3 = map(x->x + x÷2, x2) #3x
z = map(+, x, x3) #4x
```
Starts as 8 (4*2), then after `push!(x, 3)` it updates to 9, not 12 (4*3).

This worked ok before this PR (with depth first propagation).

Essentially, signals need to propagate in the order they are created, which will always be a topological order since you can't create a signal (map, filter, etc.) before it's input signals have been created.

PR #121, opened 2017-01-31
Allows naming signals

JobJob:
Useful for debugging and could be used for visualising signal graphs
They will get an auto-incremented node-number as their name if none is provided.

Aside: I'm a little busy right now so if you want to make changes yourself that would be so lovely :D

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -1,5 +1,5 @@
 import Base: push!, eltype, close
-export Signal, push!, value, preserve, unpreserve, close
+export Signal, push!, value, preserve, unpreserve, close, rename!
 
 using DataStructures
 
@@ -17,6 +17,7 @@ if !debug_memory
         actions::Vector
         alive::Bool
         preservers::Dict
+        name::String
     end
 else
     type Signal{T}
@@ -25,9 +26,10 @@ else
         actions::Vector
         alive::Bool
         preservers::Dict
+        name::String
         bt
-        function Signal(v, parents, actions, alive, pres)
-            n=new(v,parents,actions,alive,pres,backtrace())
+        function Signal(v, parents, actions, alive, pres, name)
+            n=new(v,parents,actions,alive,pres,name, backtrace())
             nodes[n] = nothing
             finalizer(n, log_gc)
             n
@@ -49,14 +51,32 @@ immutable Action
     f::Function
 end
 
+const node_count = DefaultDict{String,Int}(0) #counts of different signals for naming
+function auto_name!(name, parents...)
+    parents_str = join(map(s->s.name, parents), ", ")
+    isempty(parents_str) || (name *= "($parents_str)")
+    node_count[name] += 1
+    countstr = node_count[name] > 1 ? "-$(node_count[name])" : ""
+    "$name$countstr"
+end
+
+"""
+`rename!(s::Signal, name::String)`
+
+Change a Signal's name
+"""
+function rename!(s::Signal, name::String)
+    s.name = name
+end
+
 const action_queue = Queue(Tuple{Signal, Action})
 
 isrequired(a::Action) = (a.recipient.value != nothing) && a.recipient.value.alive
 
-Signal{T}(x::T, parents=()) = Signal{T}(x, parents, Action[], true, Dict{Signal, Int}())
-Signal{T}(::Type{T}, x, parents=()) = Signal{T}(x, parents, Action[], true, Dict{Signal, Int}())
+Signal{T}(x::T, parents=(); name::String=auto_name!("input")) = Signal{T}(x, parents, Action[], true, Dict{Signal, Int}(), name)
+Signal{T}(::Type{T}, x, parents=(); name::String=auto_name!("input")) = Signal{T}(x, parents, Action[], true, Dict{Signal, Int}(), name)
 # A signal of types
-Signal(t::Type) = Signal(Type, t)
+Signal(t::Type; name::String = auto_name!("input")) = Signal(Type, t, name)
 
 # preserve/unpreserve nodes from gc
 """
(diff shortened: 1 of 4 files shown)


JobJob:
Oh, looks like I only implemented this for Signal and map... Don't really have time to do the others right now 😬

JobJob:
Doing a bit of cleanup. Tests are failing.

src/core.jl
```diff
@@ -9,6 +9,9 @@ const debug_memory = false # Set this to true to debug gc of nodes
 
 const nodes = WeakKeyDict()
 const io_lock = ReentrantLock()
+const node_count = Ref(1) #1 since auto_name gets called before count is incremented
+
+auto_name() = string(node_count[])
```
shashi: maybe `string("signal-", node_count[])`?

JobJob:
Ok updated and (hastily) added the name option to most operators.

N.b. API for `onerror` callback is changed node where error occurred also passed to the callback now.

src/core.jl
shashi: increment should also be done here.

src/core.jl
```diff
@@ -17,6 +20,11 @@ if !debug_memory
         actions::Vector
         alive::Bool
         preservers::Dict
+        name::String
+        function Signal(v, parents, actions, alive, pres, name)
+            node_count[] += 1
+            n=new(v,parents,actions,alive,pres,name)
+        end
```
shashi: Can remove this if increment is moved to `auto_name`

src/operators.jl
```diff
@@ -235,8 +236,8 @@ Flatten a signal of signals into a signal which holds the
 value of the current signal. The `typ` keyword argument specifies
 the type of the flattened signal. It is `Any` by default.
 """
-function flatten(input::Signal; typ=Any)
-    n = Signal(typ, value(value(input)), (input,))
+function flatten(input::Signal; typ=Any, name=auto_name())
+    n = Signal(typ, value(value(input)), (input,); name=name)
```
shashi: I wonder if instead of adding keyword arguments to everything we can define `name!(signal, "name")` to set the name for signals you want to name.

It might also be nicer to default have default names such as "map(signal-1, signal-2)-1", "filter(signal-1)-2" etc.

JobJob: Great ideas, but I've no time right now. 
Add them later?

shashi:
@JobJob Thanks! no particular hurry here. I'll have a look at this later and make the changes.

JobJob:
Alright great, please do, thanks.

src/operators.jl
JobJob: I think naming them on creation makes sense, and the kw args are already there now. I've also added a `rename!(s::Signal, name::String)`

src/core.jl
JobJob: yep - done

JobJob:
Added a new commit, I think it's probably ready.

Dug this out from a comment on an old commit, since it's the one remaining discussion point:
> I wonder if instead of adding keyword arguments to everything we can define name!(signal, "name") to set the name for signals you want to name.

Auto-naming them on creation makes sense to me, and the kw args are already there now. Any good reasons to change it to explicit `name!`? What would the default be in that case - empty string?

Also I've added a `rename!(s::Signal, name::String)`

> It might also be nicer to default have default names such as "map(signal-1, signal-2)-1", "filter(signal-1)-2" etc.

This is implemented now.

src/core.jl
```diff
@@ -17,6 +17,10 @@ if !debug_memory
         actions::Vector
         alive::Bool
         preservers::Dict
+        name::String
+        function Signal(v, parents, actions, alive, pres, name)
+            n=new(v,parents,actions,alive,pres,name)
```
shashi: what's the use of this constructor?

shashi:
Sweet!

> Any good reasons to change it to explicit `name!`?

Just to avoid having an extra kwarg to maintain for every function. But I'm not crazy about it though. This looks good!

src/core.jl
JobJob: No use, maybe it's vestigial from an earlier version where I had autoname in there? can't remember, either way, let's remove it

src/operators.jl
```diff
@@ -24,9 +24,10 @@ export map,
 Transform signal `s` by applying `f` to each element. For multiple signal arguments, apply `f` elementwise.
 """
 function map(f, input::Signal, inputsrest::Signal...;
```
shashi: I think because it was causing ambiguity for the `map(f)` method sinc Base defines `map(f, xs...)` too.

JobJob:
I can't merge this btw, don't have [the power](https://www.youtube.com/watch?v=V8h8snfYidg). I will squash the commits.

src/operators.jl
JobJob: ok great, thanks.

JobJob:
Squashed.

shashi:
huh that's weird what's the point of an org member then. I've invited you as a collaborator on this repo.

tknopp:
@shashi: I think you can selectively control the repositories for an org member (with the default of no access) whereas a collaborator has access to every repo

src/core.jl
```diff
@@ -223,7 +244,7 @@ function run(n::Int=typemax(Int))
                     rethrow()
                 else
                     bt = catch_backtrace()
-                    msgval.onerror(msgval.node, msgval.value, CapturedException(err, bt))
+                    msgval.onerror(msgval.node, msgval.value, node, CapturedException(err, bt))
```
timholy: The docstring for `push!` needs to be updated to specify the new arguments for `onerror`. This is a breaking change (the next tag will have to be 0.4.0), as anyone using `onerror` now needs to change their callback functions. Unfortunately, I don't see how one can emit a deprecation warning, except possibly to look for this in failures of the `try`. It also needs to be advertised somehow, e.g., in a `NEWS.md` file.

src/operators.jl
```diff
@@ -143,9 +145,9 @@ Merge many signals into one. Returns a signal which updates when
 any of the inputs update. If many signals update at the same time,
 the value of the *youngest* input signal is taken.
 """
-function merge(inputs...)
+function merge(inputs...; name=auto_name!("merge", inputs...))
```
tkelman: is this missing argument types? can this avoid pirating Any... inputs?

Issue #123, opened 2017-03-02
`throttle` signal not updated


skleinbo:
Reactive: 0.3.7
Julia: 0.5.0

Maybe I am misunderstanding how `throttle` works, but I would expect the following code to print a line every second

```julia
using Reactive

x=fps(15)
y=throttle(1.0, x)
preserve(map(println, y))
```

Nothing happens though. `y` is not updated with `x`. Note that printing `x` directly works perfectly.

So, what's going on, or what is it I do not understand?

Thank you!

JobJob:
The current implementation of `throttle` is that it won't "fire" until `dt` (first argument) seconds have passed since it was last "pushed" to. So, what's happening is, since `x` never stops "pushing" to it (it's pushing every ~1/15 secs), 1.0 secs never goes by before it gets "pushed" to again, which restarts the 1.0 second timer, and thus it never "fires". 

It's more like underscore.js's [debounce](http://underscorejs.org/#debounce) if that helps.

I'm not sure whether that's ideal, but that's the current state of play. There might well be another way to achieve what you want with some combination of `fps(1.0)`,  `sampleon` and possibly `foldp` if you want to give more details.

skleinbo:
@JobJob 
Thank you. That clears it up for me! I would still argue that `throttle` is a bit misleading. And so is the doctring.

But for my immediate usecase, combining `fps` and `sampleon` will do perfectly fine.

May be closed.

JobJob:
You now have a throttle that behaves as you would expect on master. The old throttle behaviour is retained in `debounce`. See `?throttle` and `?debounce` for more info.

Issue #119, opened 2017-01-26
BindedTo


TsurHerman:
I find it would be handy to have a BindedTo method returning all the bindings of a signal..

TsurHerman:
I don't want to swamp the issues here .. but I found a disturbing bug
```

julia> using Reactive

julia> x = Signal(0)
Signal{Int64}(0, nactions=0)

julia> y = Signal(1)
Signal{Int64}(1, nactions=0)

julia> bind!(y,x,false)

julia> push!(x,10)

julia> value(y)
10

julia> unbind!(y,x,true)

julia> gc()

julia> push!(x,100)

julia> value(y)
100

julia>
```

unbind! doesn't work :(

JobJob:
You now have `bound_srcs` and `bound_dests` on master. e.g.
```
Pkg.checkout("Reactive")
using Reactive
x = Signal(0)
y = Signal(1)
bind!(y,x,false)
@show bound_srcs(y)[1] == x
@show bound_dests(x)[1] == y
```

We can close this yeah?

PR #129, opened 2017-03-24
Node Creation Order Redesign

JobJob:
Finally got around to cleaning this branch up.

This PR:

1. Gets signals updating in ~~the~~ a correct (topological) order (the order Signals are created), which fixes bugs in signals updating on certain signal graphs. See the last three test sets [here](https://github.com/JobJob/Reactive.jl/blob/action_queues_redesign/test/node_order.jl) for examples of where the current (breadth-first) and pre #113 (depth-first) designs had bugs.
2. Makes async optional through `Reactive.run_async(false)` - though it's not well tested, so async is on by default.
3. Runs faster than current master on @SimonDanisch's benchmark - included in https://github.com/JobJob/Reactive.jl/blob/action_queues_redesign/benchmark/ReactBench.jl - and slightly (though not much) faster in synchronous mode. Is now around 2.5x slower than the non-reactive version on my machine (previous master was about 4x slower).
4. Adds a PkgBenchmark - based off Shashi's speedups_maybe branch.
5. Fixes a bug in error reporting.
6. Adds a small amount of docs about the [design](https://github.com/JobJob/Reactive.jl/blob/action_queues_redesign/doc/Design%20Overview.md) and the [garbage collection/preserve](https://github.com/JobJob/Reactive.jl/blob/action_queues_redesign/doc/dev%20notes.md) mechanism.
7. adds `bound_srcs`, `bound_dests` see OP of #119 
8. Adds `debounce` - which behaves as per previous throttle
9. Fixes `throttle` issue in #123, but note this changes the behaviour of throttle, use debounce if you want the previous behaviour.

Downsides:

1. ~~Can't `push!` to non input nodes - those created with `Signal(...)`~~ you can now
1. ~~`flatten`~~ `bind` implementation is a little bit more complicated, though may be more correct in some cases.
1. May have bugs that aren't tested for in the tests, since the design is a little different.
1. ~~Haven't checked if it plays nice with Garbage Collection~~

Todo:

- [x] Fix interaction with Garbage Collection
- [x] Clean up some references in comments to action_queue
- [x] Add NEWS.md entry about throttle.
- [x] Fix `debug_memory` code
- [x] Make throttle tests more robust
- [x] Fix Travis test failures

### DIFF
--- a/doc/Design Overview.md
+++ b/doc/Design Overview.md
@@ -1,0 +1,41 @@
+### Node Creation Order Design
+
+When a node is `push!`ed to in user code, this library must process it and ensure signal values stay consistent with the operations users used to define the chain of signals (e.g. map, foldp, etc.).
+
+(N.b. "node" and "Signal" are used interchangeably in this doc and the code)
+
+The design assumes:
+
+1. The order which nodes are created is a correct [topological ordering](https://en.wikipedia.org/wiki/Topological_ordering) (with the edges of the signal graph  regarded as directed from parents to children)
+2. Signals will end up in a correct state if the order in which each node is processed and their update actions (e.g. the mapped function in the case of a map) run, is the same as the order in which nodes were created.
+3. Signal actions should be run for a given `push!` only if the node itself was pushed to or if one of their parents had their actions run.
+
+This should ensure that parents of nodes update before their children, and signal values will be in a correct state after each `push!` has been processed.
+
+#### Basics
+
+Each node (`Signal`) is added to the end of a Vector called `nodes` on creation, so that `nodes` holds Signals in the order they were created.
+
+Each Signal holds a field `actions` which are just 0-argument functions that update the value of the node or perform some helper function to that end. In some cases the action will update, push to, or set a Timer to update a different node.
+
+Each Signal also has a field `active` which flags whether or not the node was `push!`ed to, or had/should have its actions run, in processing the current push. In essence it flags whether or not the node's value has been updated, or should be updated.
+
+Nodes that are pushed to will always be set to active, other (downstream) nodes will be set to active, and their actions run if any of their parent `Signal`s were active in processing the current `push!`.
+
+On processing each `push!`, we run through `nodes` and execute the actions of each node if it has been set to active, i.e. if it was pushed to, or if any of its parents were active.
+
+#### Pushing to Non Input Nodes
+
+Sometimes it is desirable to push! a value to a non-input node, e.g. a `map` on an input `Signal(...)`, rather than it attaining that value by running its action. In order for this pushed value to "stick", it's important that the map's action does not run after pushing to the node - since the map's action would update the map node to the return value of the function used to create the map, which in general would not be equal to the pushed value.
+
+This is achieved simply by the check in `run_node` requiring an active parent in order to run the node's actions.
+
+A consequence of this is any actions attached to a node with no parents, e.g. an input `Signal(...)` node, will not run. Accordingly, all actions that rely on an update to a node, are attached to a child of the node, and not the node itself. See [dev notes](dev%20notes.md) for more details.
+
+#### Filter
+
+Filter works by setting the filter node's active field to false when the filter condition is false. Downstream/descendent nodes check if at least one of their parents has been active, if none of them have been active then the node will not run its action, thus propagating the filter correctly.
+
+#### More info
+
+There is some info on each operator in the [dev notes](dev%20notes.md). Please feel free to open issues if things are not clear.
--- a/src/core.jl
+++ b/src/core.jl
@@ -7,35 +7,57 @@ using DataStructures
 
 const debug_memory = false # Set this to true to debug gc of nodes
 
-const nodes = WeakKeyDict()
+const nodeset = WeakKeyDict()
 const io_lock = ReentrantLock()
 
 if !debug_memory
     type Signal{T}
+        id::Int # also its index into `nodes`, and `edges`
         value::T
         parents::Tuple
-        actions::Vector
-        alive::Bool
+        active::Bool
+        actions::Vector{Function}
         preservers::Dict
         name::String
+        function (::Type{Signal{T}}){T}(v, parents, pres, name)
+            id = length(nodes) + 1
+            n=new{T}(id, v, parents, false, Function[], pres, name)
+            push!(nodes, WeakRef(n))
+            push!(edges, Int[])
+            foreach(p->push!(edges[p.id], id), parents)
+            finalizer(n, schedule_node_cleanup)
+            n
+        end
     end
 else
     type Signal{T}
+        id::Int
         value::T
         parents::Tuple
+        active::Bool
         actions::Vector
-        alive::Bool
         preservers::Dict
         name::String
         bt
-        function (::Type{Signal{T}}){T}(v, parents, actions, alive, pres, name)
-            n=new{T}(v,parents,actions,alive,pres,name, backtrace())
-            nodes[n] = nothing
+        function (::Type{Signal{T}}){T}(v, parents, actions, pres, name)
+            id = length(nodes) + 1
+            n=new{T}(id, v, parents, false, Function[], pres, name, backtrace())
+            push!(nodes, WeakRef(n))
+            push!(edges, Int[])
+            foreach(p->push!(edges[p.id], id), parents)
+            nodeset[n] = nothing
             finalizer(n, log_gc)
             n
         end
     end
 end
+
+Signal{T}(x::T, parents=(); name::String=auto_name!("input")) =
+    Signal{T}(x, parents, Dict{Signal, Int}(), name)
+Signal{T}(::Type{T}, x, parents=(); name::String=auto_name!("input")) =
+    Signal{T}(x, parents, Dict{Signal, Int}(), name)
+# A signal of types
+Signal{T}(t::Type{T}; name::String = auto_name!("input")) = Signal(Type{T}, t, name=name)
 
 log_gc(n) =
     @async begin
@@ -46,10 +68,9 @@ log_gc(n) =
         unlock(io_lock)
     end
 
-immutable Action
-    recipient::WeakRef
-    f::Function
-end
+const nodes = WeakRef[] #stores the nodes in order of creation (which is a topological order for execution of the nodes' actions)
+
+const edges = Vector{Int}[] #parents to children, useful for plotting graphs
 
 const node_count = DefaultDict{String,Int}(0) #counts of different signals for naming
 function auto_name!(name, parents...)
@@ -68,15 +89,6 @@ Change a Signal's name
 function rename!(s::Signal, name::String)
     s.name = name
 end
-
-const action_queue = Queue(Tuple{Signal, Action})
-
-isrequired(a::Action) = (a.recipient.value != nothing) && a.recipient.value.alive
-
-Signal{T}(x::T, parents=(); name::String=auto_name!("input")) = Signal{T}(x, parents, Action[], true, Dict{Signal, Int}(), name)
-Signal{T}(::Type{T}, x, parents=(); name::String=auto_name!("input")) = Signal{T}(x, parents, Action[], true, Dict{Signal, Int}(), name)
-# A signal of types
-Signal{T}(t::Type{T}; name::String = auto_name!("input")) = Signal(Type{T}, t, name = name)
 
 # preserve/unpreserve nodes from gc
 """
@@ -111,8 +123,10 @@ function unpreserve(x::Signal)
     x
 end
 
-Base.show(io::IO, n::Signal) =
-    write(io, "$(n.name): Signal{$(eltype(n))}($(n.value), nactions=$(length(n.actions))$(n.alive ? "" : ", closed"))")
+Base.show(io::IO, n::Signal) = begin
+    active_str = isactive(n) ? "(active)" : ""
+    write(io, "$(n.id): \"$(n.name)\" = $(n.value) $(eltype(n)) $active_str")
+end
 
 value(n::Signal) = n.value
 value(::Void) = false
@@ -121,51 +135,62 @@ eltype{T}(::Type{Signal{T}}) = T
 
 ##### Connections #####
 
-function add_action!(f, node, recipient)
-    a = Action(WeakRef(recipient), f)
-    push!(node.actions, a)
+function add_action!(f, node)
+    push!(node.actions, f)
     maybe_restart_queue()
-    a
+    f
 end
 
-function remove_action!(f, node, recipient)
-    node.actions = filter(a -> a.f != f, node.actions)
+"""
+Removes `action` from `node.actions`
+"""
+function remove_action!(node, action)
+    filter!(a -> a != action, node.actions)
+    nothing
+end
+
+const run_remove_dead_nodes = Ref(false)
+"""
+Schedule a cleanup of dead nodes - called as a finalizer on each GC'd node
+"""
+function schedule_node_cleanup(n)
+    run_remove_dead_nodes[] = true
+end
+
+"""
+Remove GC'd nodes from `nodes`, is run before push! when scheduled.
+Not thread-safe in the sense that if it is run while other code is iterating
+through `nodes`, e.g. in run_push, iteration could skip nodes.
+"""
+function remove_dead_nodes!()
+    allnodes = [] # keep a ref to all nodes to avoid GC whilst in this function
+    filter!(nodes) do noderef
+        node = noderef.value
+        node != nothing && push!(allnodes, node)
+        node != nothing
+    end
+    foreach((i_n)->((i,n) = i_n; n.id = i), enumerate(allnodes)) # renumber nodes
+    reinit_edges!(allnodes)
+    nothing
+end
+
+function reinit_edges!(allnodes)
+    empty!(edges)
+    foreach(n->push!(edges,[]), allnodes)
+    foreach(allnodes) do n
+        foreach(p->push!(edges[p.id], n.id), n.parents)
+    end
+    nothing
 end
 
 function close(n::Signal, warn_nonleaf=true)
-    finalize(n) # stop timer etc.
-    n.alive = false
-    if !isempty(n.actions)
-        any(map(isrequired, n.actions)) && warn_nonleaf &&
-            warn("closing a non-leaf node is not a good idea")
-        empty!(n.actions)
-    end
-
     for p in n.parents
         delete!(p.preservers, n)
     end
+    finalize(n) # stop timer, schedule_node_cleanup, etc.
 end
 
-function send_value!(node::Signal, x, timestep)
-    # Dead node?
-    !node.alive && return
-
-    # Set the value and do actions
-    node.value = x
-    for action in node.actions
-        action.recipient.value != nothing && #nothing means downstream node has been gc'd
-            DataStructures.enqueue!(action_queue, (action.recipient.value, action))
-    end
-end
-send_value!(wr::WeakRef, x, timestep) = wr.value != nothing && send_value!(wr.value, x, timestep)
-
-do_action(a::Action, timestep) =
-    isrequired(a) && a.f(a.recipient.value, timestep)
-
-# If any actions have been gc'd, remove them
-cleanup_actions(node::Signal) =
-    node.actions = filter(isrequired, node.actions)
-
+set_value!(node::Signal, x) = (node.value = x)
 
 ##### Messaging #####
 
@@ -180,6 +205,9 @@ end
 # Global channel for signal updates
 const _messages = Channel{Nullable{Message}}(CHANNEL_SIZE)
 
+#run in asynchronous mode by default
+const async_mode = Ref(true)
+run_async(async::Bool) = (async_mode[] = async)
 
 """
 `push!(signal, value, onerror=Reactive.print_error)`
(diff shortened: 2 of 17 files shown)


shashi:
Exciting!

SimonDanisch:
Nice! :)
Can we avoid downside 1?! I'm not heavily relying on it, but I'm using it here and there ;)

JobJob:
> Can we avoid downside 1?! I'm not heavily relying on it, but I'm using it here and there ;)

Will be hard to do it cleanly now I think, will be so much better if you can live without it ;) I've mostly used up the time budget I allocated to cleaning this up 😬 

When I was thinking about this issue, I figured that if needed, people could just `bind!` a new Signal to the one they previously wanted to `push!` to and use that instead, it would then function like the original Signal, but you could also `push!` to that new Signal when you wanted to.

JobJob:
Is that ok? I'll give it a little more thought, it might be doable, but I'd much prefer not to 😄  I am curious about the situations you want to do it, I guess I'd be more inclined to looking into it if it makes doing some relatively common things much easier.

SimonDanisch:
Okay fair enough ;) My use case is that I have my API designed also for people who don't want to use Reactive:
```Julia
visualization = visualize(..., color = user_signal) # user_signal might also be a default, non Signal node created by me)
set_arg!(visualization, :color, newvalue) # should push into the signal, regardless of what got there
```
Ah well, I'll need a redesign anyways...

JobJob:
I thought I'd quickly whip up an alternative for you @SimonDanisch but I found a bug in `bind!`, the fixes are now pushed. I had to disable two-way bind.

I've also pushed the branch (action_queues_redesign) to this (JuliaGizmos) repo, to make it easier to try out.

Anyway here's one way you could do what (I think) you want using `bind!`:
```
using Reactive, Interact, Colors

immutable Viz
    color::Signal
end
r,g,b = Signal.([0.7, 0.1, 0.3]) #create r,g and b signals
default_color_sig = 
    map((rv,gv,bv)->RGB(rv,gv,bv), r,g,b; name="default_color_sig") #default sig is a map

visualization(user_signal = default_color_sig) = begin
    color_sig = Signal(colorant"cyan"; name="pushable_color_sig") #can also be pushed to
    bind!(color_sig, user_signal) #if user_signal is updated, update the color_sig
    Viz(color_sig)
end

set_arg!(viz, sym, val) = push!(getfield(viz, sym), val) #will work with a regular value
the_viz = visualization()
display(the_viz.color);
```

then you can update it with e.g.
```
set_arg!(the_viz, :color, colorant"pink")
```
or
```
push!(g, 0.9)
```

JobJob:
I think I can actually allow pushing to non-input nodes in a relatively clean way. Will confirm shortly.

JobJob:
Ok yes, `push!` works again to non-input nodes, and I managed to restore the previous two way bind behaviour, so all looking pretty good. Will check the gc stuff in the next couple of days.

src/operators.jl
```diff
@@ -33,15 +35,9 @@ function map(f, input::Signal, inputsrest::Signal...;
...
-    let prev_timestep = 0
-        for inp in inputs
-            add_action!(inp, output) do output, timestep
-                if prev_timestep != timestep
-                    result = f(map(value, inputs)...)
-                    send_value!(output, result, timestep)
-                    prev_timestep = timestep
-                end
-            end
+    let
```
timholy: Do you still need the `let`? (Here and elsewhere.)

src/operators.jl
```diff
@@ -244,65 +267,161 @@ function flatten(input::Signal; typ=Any, name=auto_name!("flatten", input))
...
+                #ensure updates when current_node gets pushed a new value
+                add_action!(output, root) do output
+                    send_value!(output, value(current_node))
+                end
+                #ensure updates for subtree too, if action is already in queue
+                #remove it and readd it so it's deeper down. XXX I think that's
+                #mostly correct but prob not 100%
+                queue = action_queues[root]
+                for action in subtree_actions
+                    deleteat!(queue, findin(queue, 6))
```
timholy: What's the 6 about?

src/operators.jl
```diff
@@ -244,65 +267,161 @@ function flatten(input::Signal; typ=Any, name=auto_name!("flatten", input))
...
-        action = add_action!(b, a) do a, timestep
-            if current_timestep != timestep
-                current_timestep = timestep
-                send_value!(a, value(b), timestep)
-            end
+Note that signals dependent on `dest` added before calling `bind!` (e.g. those
+created by calling `map(f, dest)`, `merge(dest, s1)`, `filter(f, dest)`, etc.
+before `bind!(src, dest)`) may be updated more than once when `src` updates. To
+ensure signals update just once per `push!`, call `bind!` before adding signals
+dependent on `dest`.
```
timholy: Was this also an issue with the former implementation?

src/operators.jl
```diff
@@ -244,65 +267,161 @@ function flatten(input::Signal; typ=Any, name=auto_name!("flatten", input))
...
 
-    let current_timestep = 0
-        action = add_action!(b, a) do a, timestep
-            if current_timestep != timestep
-                current_timestep = timestep
-                send_value!(a, value(b), timestep)
-            end
+Note that signals dependent on `dest` added before calling `bind!` (e.g. those
+created by calling `map(f, dest)`, `merge(dest, s1)`, `filter(f, dest)`, etc.
+before `bind!(src, dest)`) may be updated more than once when `src` updates. To
```
timholy: Check argument order to `bind!`

doc/action queue design.md
```diff
@@ -0,0 +1,17 @@
...
+
+Each node uses the field `roots` to store the keys (nodes) of the action_queues they are in.
+
+#### Filter
+
+Filter works by setting the filter node's active field to false when the filter condition is false. Downstream nodes (nodes later in the current `action_queue`) check if at least one of their parents has been active, if none of them have been active then the node will not run it's action, thus propagating the filter correctly.
+
+#### Flatten
+
+Implementation is a bit fiddly. See comments in the code for how the graph is rewired.
```
JobJob: Could definitely flesh that doc out a bit, but low priority for now. The docstring for flatten has a bit of info. Please do let me know if you have questions about the implementation though.

src/core.jl
```diff
@@ -14,29 +14,70 @@ if !debug_memory
...
+empty, rather than having `node.roots == (node,)` to avoid an unecessary
+self-reference - which would require working around problems with printing, saving
+to disk etc. Instead, wherever we need the roots of a `node` to include the
+root node itself we call `allroots(node)`.
+"""
+allroots(maybe_root_node) =
+    isempty(maybe_root_node.roots) ? [maybe_root_node] : maybe_root_node.roots
+
+function getroots(parents)
+    roots = Dict{Signal, Bool}()
```
JobJob: I don't know why I did that, I like `Set`s, maybe I'll remember a reason...

doc/action queue design.md
```diff
@@ -0,0 +1,17 @@
+### Action Queue Design
+
+Every time a node is pushed to a sequence of actions takes place. This action sequence can, for the most part, be determined as the signal graph is being created. So for each (root/input) node, a vector of actions - an `action_queue` - is set up and stored in `action_queues[node]`. When `node`s are created that depend on other nodes, they are added to all the appropriate action queues, i.e. all the action_queues that the parent nodes are in.
+
+Each node uses the field `roots` to store the keys (nodes) of the action_queues they are in.
+
+#### Filter
+
+Filter works by setting the filter node's active field to false when the filter condition is false. Downstream nodes (nodes later in the current `action_queue`) check if at least one of their parents has been active, if none of them have been active then the node will not run it's action, thus propagating the filter correctly.
```
JobJob: Ta - once!

src/core.jl
```diff
@@ -121,50 +168,51 @@ eltype{T}(::Type{Signal{T}}) = T
...
 
-function close(n::Signal, warn_nonleaf=true)
-    finalize(n) # stop timer etc.
-    n.alive = false
-    if !isempty(n.actions)
-        any(map(isrequired, n.actions)) && warn_nonleaf &&
-            warn("closing a non-leaf node is not a good idea")
-        empty!(n.actions)
+"""
+remove the action associated with the node from all of it's roots' action_queues
```
JobJob: Twice! :/

src/operators.jl
JobJob: I don't think so. Didn't want to mess with closures around the action functions where locals are defined, but I think you're right that they'd probably still work without the `let`s

src/operators.jl
JobJob: lol I have no idea! Need to work that out, how does this even work!?

src/operators.jl
JobJob: 👍

src/operators.jl
JobJob: I don't think so, this was a change late in the day, I should add it to the downsides on the PR notes once I've checked it. 

Note this is only a problem if someone does the thing that the little note says not to do, and has a downstream op that has some side-effect that they don't want running more than once per-push, or perhaps some expensive op that would run twice and thus impact performance.

src/operators.jl
JobJob: I think it's supposed to be `action`. I guess it worked but would have caused some things to be re-run twice since in some cases since they weren't removed properly.

JobJob:
After spending some time with the code in this branch again, trying to fix the `bind!` issue, I've come to the view that I don't like the design. It's too complicated, too much pain, not enough gain. I have an idea of how to simplify it. Will flesh it out tomorrow, I hope I can at least salvage some of it. Changed to WIP in the interim.

JobJob:
Simpler design is coming along well so far. Will update over the weekend.

JobJob:
Everything's looking pretty good I think on the simpler design. Tests are passing, managed to restore the old `bind!` behaviour, I think it's about the same speed as my earlier more complicated version (removed some WeakRefs temporarily, so not 100%), and there's less code. Will do some cleanup and push here tomorrow.

JobJob:
Pushed my latest, its running reasonably fast, about the same as the earlier version, possibly faster, but its using fewer WeakRefs which may affect things slightly, since I haven't fixed up how it interacts with Garbage Collection yet. Will do that tomorrow.

There was talk about changing how Reactive behaves with Garbage collection, if someone has some ideas, now would be a good time, otherwise I'll just restore the previous behaviour.

There's a new, still minimal, design doc here: https://github.com/JuliaGizmos/Reactive.jl/blob/action_queues_redesign/doc/Design%20Overview.md

JobJob:
I think this is pretty much ready now

doc/Design Overview.md
```diff
@@ -0,0 +1,31 @@
...
+
+Each Signal holds a field `actions` which are basically just 0-argument functions that update the value of the node or perform some helper function to that end.
+
+Each Signal also has a field `active` which stores whether or not the node was active (had its actions run) in the current push. A Signal will be set to active if it is `push!`ed to, or if any of its parent `Signal`s were active.
+
+On processing each `push!`, we run through `nodes` and execute the actions of each node if any of its parents were active.
+
+#### Filter
+
+Filter works by setting the filter node's active field to false when the filter condition is false. Downstream/descendent nodes check if at least one of their parents has been active, if none of them have been active then the node will not run its action, thus propagating the filter correctly.
```
shashi: neat.

doc/Design Overview.md
```diff
@@ -0,0 +1,31 @@
...
+
+#### Basics
+
+Each node (`Signal`) is added to the end of a Vector called `nodes` on creation, so that `nodes` holds Signals in the order they were created.
+
+Each Signal holds a field `actions` which are basically just 0-argument functions that update the value of the node or perform some helper function to that end.
+
+Each Signal also has a field `active` which stores whether or not the node was active (had its actions run) in the current push. A Signal will be set to active if it is `push!`ed to, or if any of its parent `Signal`s were active.
+
+On processing each `push!`, we run through `nodes` and execute the actions of each node if any of its parents were active.
```
shashi: I like the simplicity of this design very much. The global channel makes sure that two push!es don't overlap, so this seems correct.

test/flatten.jl
```diff
@@ -47,4 +43,49 @@ facts("Flatten") do
         @fact value(cnt) --> 3
         @fact value(d) --> value(b)
     end
+
+    context("Subtle sig swap issue") do
+        # When a node, a map (e) in this case, has a flatten as a parent, but also
+        # a signal that is the flatten parent sigsig's (c's) current value ("a" here)
+        # then when the sigsig gets pushed another value, you want the map to
+        # still update on changes to a, even after the map is "rewired" when a
+        # new value is pushed to c.
```
shashi: :+1:

test/node_order.jl
```diff
@@ -0,0 +1,154 @@
...
+    sx = Signal(xv)
+    s1x1 = map(identity, sx)
+    s1x2 = map(identity, sx)
+    s2x = map(+, s1x1, s1x2)
+    s3x = map(+, sx, s2x)
+    @fact value.([sx, s1x1, s1x2, s2x, s3x]) --> [2, 2, 2, 4, 6]
+    push!(sx, 3)
+    Reactive.run_till_now()
+    @fact value.([sx, s1x1, s1x2, s2x, s3x]) --> [3, 3, 3, 6, 9]
+end
```
shashi: victory!

src/core.jl
```diff
@@ -7,36 +7,58 @@ using DataStructures
...
-        alive::Bool
         preservers::Dict
         name::String
         bt
-        function (::Type{Signal{T}}){T}(v, parents, actions, alive, pres, name)
-            n=new{T}(v,parents,actions,alive,pres,name, backtrace())
-            nodes[n] = nothing
+        function (::Type{Signal{T}}){T}(v, parents, actions, pres, name)
+            id = length(nodes) + 1
+            n=new{T}(id, v, parents, false, Action[], pres, name, backtrace())
```
JobJob: That seems like a really good idea!

src/core.jl
```diff
@@ -46,11 +68,15 @@ log_gc(n) =
...
     end
 
+# TODO remove this, don't need recipient anymore so can just change all to Functions
 immutable Action
     recipient::WeakRef
     f::Function
 end
 
+const nodes = WeakRef[] #stores the nodes in order of creation (which is a topological order for execution of the nodes' actions)
+const edges = Vector{Int}[] #parents to children, useful for plotting graphs
```
JobJob: Yeah, also check this:
```
using Reactive, LightGraphs, Plots, PlotRecipes
pyplot()

s = Signal(1; name="sig 1")
m2 = map(x->2x, s; name="x2")
m3 = map(x->3x, s; name="x3")

function LightGraphs.DiGraph{T <: Integer}(edgelists::Vector{Vector{T}})
    g = DiGraph(length(edgelists))
    for (v1, edgelist) in enumerate(edgelists)
        foreach(v2->add_edge!(g, v1, v2), edgelist)
    end
    g
end

g = DiGraph(Reactive.edges)
graphplot(g, arrow = arrow(:simple, :head, 1.5, 1.0) , names=map(nr->nr.value.name, Reactive.nodes), linewidth=5)
```

src/core.jl
```diff
@@ -121,51 +140,40 @@ eltype{T}(::Type{Signal{T}}) = T
 
 ##### Connections #####
 
-function add_action!(f, node, recipient)
-    a = Action(WeakRef(recipient), f)
+function add_action!(f, node)
+    a = Action(WeakRef(node), f)
```
JobJob: Yeah, I was planning to do that. `action.recipient` isn't used anymore

src/core.jl
```diff
@@ -121,51 +140,40 @@ eltype{T}(::Type{Signal{T}}) = T
...
-send_value!(wr::WeakRef, x, timestep) = wr.value != nothing && send_value!(wr.value, x, timestep)
-
-do_action(a::Action, timestep) =
-    isrequired(a) && a.f(a.recipient.value, timestep)
-
-# If any actions have been gc'd, remove them
-cleanup_actions(node::Signal) =
-    node.actions = filter(isrequired, node.actions)
-
+send_value!(node::Signal, x) = (node.value = x)
```
JobJob: `set_value!` could be useful to isolate the state change. I imagine it would get inlined everywhere and not affect performance, but I'll check.

src/core.jl
```diff
@@ -210,72 +227,102 @@ function _push!(n, x, onerror=print_error)
...
-    show(io, value)
+    show(io, val)
     println(io)
     println(io, "to node")
     print(io, "    ")
-    show(io, node)
+    show(io, pushnode)
     println(io)
     println(io)
     println(io, "error at node: $error_node")
```
JobJob: Yep

src/operators.jl
```diff
@@ -201,8 +213,9 @@ function delay{T}(input::Signal{T}, default=value(input); name=auto_name!("delay
 end
 
 function connect_delay(output, input)
-    add_action!(input, output) do output, timestep
-        push!(output, value(input))
+    add_action!(output) do
+        # only push when input is active (avoids it pushing to itself endlessly)
+        input.active && push!(output, value(input))
```
JobJob: You're right. I haven't given much thought to what will break in synchronous case. Will think about it.

test/node_order.jl
JobJob: Thanks for the review!

src/core.jl
shashi: Yes, just changing the name to `set_value` would be an improvement I think.

JobJob: `set_value` or `set_value!`?

src/core.jl
```diff
@@ -210,72 +227,102 @@ function _push!(n, x, onerror=print_error)
     end
     nothing
 end
-_push!(::Void, x, onerror=print_error) = nothing
-
-const timestep = Ref{Int}(0)
+async_push!(::Void, x, onerror=print_error) = nothing
```
JobJob: Actually, that line won't run because we only define `push!(n::Signal, ...`. do you think we should define `push!(::Void, ... ) = nothing` ?

src/core.jl
JobJob: Removed Actions, just using Functions now

JobJob:
Tests are failing on Travis, I'm guessing GC related, they work on my local. Will check it out tomorrow.

src/core.jl
shashi: Cool :)

JobJob:
While I was looking at GC issues, I noticed pushing to non-input nodes wasn't working, and as a result `bind!` for non-input nodes wasn't either. So since your review @shashi I had to make some changes to bind (it's now simpler), throttle, delay, fpswhen and a minor change to flatten (slightly simplified). Some of those operations now use a foreach node to run their action. See https://github.com/JobJob/Reactive.jl/blob/action_queues_redesign/doc/dev%20notes.md for some details. Added tests for pushing to non-input nodes, and they're all passing now.

I also made some changes to the way GC'd nodes are removed from `nodes` it now only happens on the main thread in the push/queue running task (does GC run in a separate thread in 0.5/0.6 or just a task/green thread?). Earlier I was modifying `nodes` in the finalizer and it was messing with iteration over `nodes` on the push running task. This seems to have eliminated GC related errors.

It's running reasonably fast. Maybe you wizards can make it run even faster - varies between 2.2x to 2.8x slower than non-reactive function calls on [Simon's benchmark](https://github.com/JobJob/Reactive.jl/blob/action_queues_redesign/benchmark/ReactBench.jl) on my local. Before the changes to make it do the right thing with GC, it was closer to 2x slower.

Ok now I really think this branch is almost ready to go - just need to squash the commits and clean up the commit messages. I guess I'll make it into 2 commits? One for the previous design and one for the new one?

Anyway, I think it's simpler, faster, and more correct, so, yeah! 😄

shashi:
This is looking very good. The only minor comment I have is about `throttle` changing behavior since previous versions. Any ideas about how to possibly deprecate neatly? I think @timholy and @SimonDanisch are the only two users of `throttle` though. They've now been warned anyway ;)

It would be nice to bring this down to <= 10 meaningful commits. I understand you're short on spare time, so I'm fine with you doing whatever is the easiest option for you. Thanks for taking this up and putting in so much amazing effort! :tada:

JobJob:
> I think @timholy and @SimonDanisch are the only two users of throttle though

Ha. I suppose we could add a warning like this when throttle is called: "Throttle's behaviour has changed, see `?throttle` for more details, for the old behaviour use `debounce`. To disable this message call `Reactive.no_throttle_warn()`". Then in x number of point releases, remove the warning in throttle, and show a message when `Reactive.no_throttle_warn` is called, saying, e.g. "This function is no longer needed". What do you guys think?

> It would be nice to bring this down to <= 10 meaningful commits. I understand you're short on spare time, so I'm fine with you doing whatever is the easiest option for you.

Yep I'll squash them to some smaller number in the next few days, see how I go.

> Thanks for taking this up and putting in so much amazing effort! 🎉

Thanks to you and @timholy for the reviews on this stupidly large PR 😄  I'm really glad its worked out 🎉

JobJob:
Ok done. With all the merge commits in this (old) branch, rebase was making life difficult, so I just started afresh and used `git read-tree ...`, so just have the one commit. Too hard/no time to do anymore. I have the entire old branch in my fork [here](https://github.com/JobJob/Reactive.jl/tree/action_queues_redesign-old) for safe-keeping.

Anyway, I think it's cowabunga time! 😄

shashi:
Works for me, merging! A bit sad that this won't help you much in the dark-green-squares-in-your-github-profile game though. :stuck_out_tongue_winking_eye:

JobJob:
Haha, just one light green square - very rough - not to mention still being 5th in the number of commits leaderboard for this repo.

Regardless - party time, excellent! 🎉  🎉  🎉

timholy:
Heroes are not measured in numbers of commits! Thanks, @JobJob!

timholy:
I'm very pleased to report that for both GtkReactive and a just-submitted [rewrite of ImageView](https://github.com/timholy/ImageView.jl/pull/115), this huge transition was absolutely seamless---only one line of test code will need changing (an explicit test of how signals print in GtkReactive). Importantly, all the "real" functionality Just Worked. Really amazing work, @JobJob!

JobJob:
Great to hear!

PR #132, opened 2017-04-18
Fix world-age problems on 0.6

timholy:
Interestingly, the `wait(Reactive.runner_task)` in `test/runtests.jl` is probably something we should have had all along; there was always a chance that some of the queue would have been processed before the event loop shut down.

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -124,6 +124,7 @@ eltype{T}(::Type{Signal{T}}) = T
 function add_action!(f, node, recipient)
     a = Action(WeakRef(recipient), f)
     push!(node.actions, a)
+    maybe_restart_queue()
     a
 end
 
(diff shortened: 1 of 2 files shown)


timholy:
Just one (spurious) `fpswhen` failure.

src/core.jl
```diff
@@ -275,9 +276,16 @@ end
...
 run_till_now() = run(Base.n_avail(_messages))
 
-# A decent default runner task
-function __init__()
-    global runner_task = @async begin
-        Reactive.run()
+# Works around world age problems (see issue #131)
+function maybe_restart_queue()
+    global runner_task
+    if !istaskdone(runner_task)
```
JobJob: Out of curiosity, when might `runner_task` be done already?

timholy: Any time it's been shut down, e.g., during the Reactive tests.

timholy:
In theory one could do this with `precompile`, though it would take some care to get the types right.

timholy:
I'd like to tag a new release fairly soon, but let's wait until there's been some method for addressing https://github.com/JuliaGizmos/Reactive.jl/pull/121#discussion_r112030635. It's worth exploring whether we can exploit the `try/catch`, but if nothing else a NEWS file would be a good idea (or a note posted at the top of the README).

shashi:
I think `try/catch` which checks to see if the thrown exception was a `MethodError` would be sufficient.

JobJob:
That section of the docstring should read something like:

The callback receives 4 arguments: the first two are the Signal and value that `push!` was called with respectively, the node whose action triggered the error, and a `CapturedException` with the fields `ex` which is the original exception object, and `processed_bt` which is the backtrace of the exception.

PR #134, opened 2017-04-20
Add type constraint to merge

tkelman:
rather than defining for ::Any

also ensure foreach has at least one Signal input


timholy:
LGTM. Thanks, @tkelman!

PR #137, opened 2017-05-11
waiting on empty queue will freeze Reactive

SimonDanisch:

PR #138, opened 2017-05-14
Fixes issue when runner task is restarted during a push

JobJob:
I've run into some issues with signals not updating recently, and I was hoping #137 was going to fix them, but it hasn't seemed to. 

Thanks to the simple reproduction in https://github.com/JuliaGizmos/GtkReactive.jl/issues/32 I was able to pin it down to situations where the queue runner task is restarted from the queue runner task itself. This will happen if add_action! is called while processing a push!. E.g. if a new map/foreach is added as part of the function of another map/foreach.

E.g. after running this
```
using Reactive
queue_size() =  Base.n_avail(Reactive._messages)
a = Signal(1)
b = Signal(2)
foreach(a; init=nothing) do _
    foreach(b; init=nothing) do _
        println("b was updated")
    end
    println("added a foreach on b")
    nothing
end
```
calling `push!(a,3)` (multiple times) was not leading to the "added a foreach on b" to be printed. Then running this showed there were actions unprocessed in the queue:
```
@show queue_size()
Reactive.run_till_now()
@show queue_size()
```
shows
```
queue_size() = 3
queue_size() = 2
```
worse still calling `run_till_now()` again causes the queue to get stuck - I think on [this line](https://github.com/JuliaGizmos/Reactive.jl/blob/master/src/core.jl#L377) , waiting on itself.

Anyway this PR fixes that. There's probably a neater way to do the restart, but I personally don't have that much time to spare for it right now. It would need to handle the case of multiple `add_action!` calls while processing the same `push!`. Anyway, I've added some tests for this so if someone has a better idea they can try it out.

It the meantime, barring someone noticing any obvious problems, I think we should merge this asap, because it came up pretty fast for me when getting back to doing some stuff with Interact in the last few days, as well as in the GtkReactive bug mentioned above - so I think it's a fairly major bug.


timholy:
LGTM. Thanks for figuring this out!

JobJob:
The travis failure is the same weird issue with loading linalg libs as on #137 
Is everyone/anyone happy for me to merge this and tag a new version?

JobJob:
Oh travis seems to be fixed, but stop press, I think I've run into another bug 😢

JobJob:
Fixed it :+1:

PR #139, opened 2017-05-15
Fix Flatten when current_node is created after the flatten

JobJob:
Found a nice bug in `flatten`: When the flatten's `current_node` (the current value of the input Signal{Signal}) is created after the flatten, the flatten won't update when the current_node updates. This is because under normal circumstances, updates to that current_node will only propagate nodes created after it. So simply having that node as a parent of the flatten, as per the existing implementation, didn't work.

The neat fix is to use `bind!` which I had previously taken care to handle an equivalent case - when the src node is downstream from (created after) the destination node.

N.b. only the last commit is relevant here. The first 3 commits are from #138 which should be merged before this.


shashi:
Cool! :+1: The merged commits appear to be still here. I will squash-merge this.

JobJob:
Ok, I can rebase right now if that's easier

src/operators.jl
```diff
@@ -265,40 +265,43 @@ value of the current signal. The `typ` keyword argument specifies
...
     current_node = input.value
     wire_flatten() = begin
         # If the sigsig's value has changed update output's parents so it will
         # only update when the new current_node updates, and no longer
         # update when the previous signal updates.
         if current_node != input.value
+            unbind!(output, current_node, false)
             current_node = input.value
-            output.parents = (input, current_node)
+            bind!(output, current_node, false)
```
shashi: This reads so nice.

JobJob: 😄

PR #142, opened 2017-05-19
fixes #141

### LINKED ISSUE #141: Should bind! ellicit a push!?
Shouty title not intended (just due to syntax).

I'm using `Reactive` (with `GtkReactive`) in a GUI. To build the GUI, I needed to `bind!` two `Signals`. But almost all the `push!` events in this GUI should only come from the user pressing on buttons etc, not from building the GUI. 

Hence, I'm wondering how wrong would it would be to bind signals without eliciting an immediate push-event to the source `Signal`? If this is a too-big change that doesn't make sense, maybe you can add an argument to bind the two signals "silently"?

yakir12:


src/operators.jl
```diff
@@ -308,12 +308,13 @@ const _bindings = Dict() # XXX GC Issue? can't use WeakKeyDict with Pairs...
 const _active_binds = Dict()
 
 """
-    `bind!(dest, src, twoway=true)`
+    `bind!(dest, src, twoway=true, initial=true)`
```
JobJob: `bind!(dest, src, twoway=true; initial=true)` I guess it should have the semi-colon?

src/operators.jl
```diff
@@ -308,12 +308,13 @@ const _bindings = Dict() # XXX GC Issue? can't use WeakKeyDict with Pairs...
...
 
 for every update to `src` also update `dest` with the same value and, if
-`twoway` is true, vice-versa.
+`twoway` is true, vice-versa, if `initial` is false, `dest` and `src` are
+not the same and `dest` doesn't get updated until `src` gets updated next.
```
JobJob: Hard to word this, maybe:
```
`twoway` is true, vice-versa. If `initial` is false, `dest` will only be updated
to `src`'s value when `src` next updates, otherwise both `dest` and `src` will
take `src`'s value immediately.
```

src/operators.jl
yakir12: oops, totally right!

src/operators.jl
yakir12: I think your wording is better, but I clarified what you meant with "otherwise":
for every update to `src` also update `dest` with the same value and, if
`twoway` is true, vice-versa. If `initial` is false, `dest` will only be updated
to `src`'s value when `src` next updates, otherwise (if `initial` is true) both
`dest` and `src` will take `src`'s value immediately.

PR #146, opened 2017-05-29
fix #144

### LINKED ISSUE #144: Maybe restart disrespects a stopped event loop
This line:
https://github.com/JuliaGizmos/Reactive.jl/blob/master/src/core.jl#L389

Will restart the event loop, even if it was manually stopped. So it basically becomes impossible to stop the event loop, since it will just restart automatically at some later point.

SimonDanisch:


JobJob:
Looks good.

As you're probably the only person that uses `stop`, I don't think it's a big issue that it changes the behaviour of `stop` slightly. If it does turn out that other people exist that do use `stop`, this is probably what they want anyway. If it's not, then they can just call `break_loop` and `run_till_now` in whatever manner they please.

Might as well also remove this https://github.com/JuliaGizmos/Reactive.jl/blob/master/test/runtests.jl#L8

A test could be good as a sanity check.

PR #150, opened 2017-07-12
add fallback for non signals

SimonDanisch:
how about it? Helps when writing code for signals + non signals...


JobJob:
I have seen in the past an issue with `value` being defined in other packages (can't remember which), but I think that's a separate issue. We can cross that bridge if we come to it again. Also, @tkelman seems to be on board with this, so, I think this is good.

PR #158, opened 2017-08-29
added a modified bind! version

### LINKED ISSUE #157: bind! with a modifier
What do you think about adding a `bind!` with some modifier?
So bind the destination to a source that's been modified by some function. For (a boolean) example:
```julia
src = Signal(true)
dst = Signal(true) # doesn't really matter what value dst has now in this example
bind!(dst, !, src) # twoway is set to true
push!(src, false) # dst is now true
push!(dst, false) # src is now true
```

yakir12:
fixes #157
now bind! accepts a modifier that changes the updates between the src
and dst.
This coolness now works:
```julia
src = Signal(30.0)
dst = Signal(sind(30.0))
bind!(dst, sind, src, asind)

push!(src, 0.0)
@fact value(dst) --> 0
push!(dst, 0.5)
@fact value(src) --> roughly(30)
```

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -370,6 +370,26 @@ function bind!(dest::Signal, src::Signal, twoway=true; initial=true)
 end
 
 """
+    `bind!(dest, src2dest, src, dest2src=nothing; initial=true)`
+
+for every update to `src` also update `dest` with a modified value (using the
+function `src2dest`) and, if `dest2src` is supplied, a two-way update will hold.
+If `initial` is false, `dest` will only be updated to `src`'s modified value
+when `src` next updates, otherwise (if `initial` is true) both `dest` and `src`
+will take the respective's modiefied values immediately. Note that currently 
+this can not be `unbind!`. Also note that for a two-way bind, a false initial
+is impossible.
+"""
+function bind!(dest::Signal, src2dest::Function, src::Signal, dest2src = nothing; initial = true)
+    dest2 = map(src2dest, src)
+    bind!(dest, dest2, false, initial = initial)
+    if dest2src ≠ nothing
+        src2 = map(dest2src, dest)
+        bind!(src2, src, true, initial = initial)
+    end
+end
+
+"""
     `unbind!(dest, src, twoway=true)`
 
 remove a link set up using `bind!`
--- a/test/basics.jl
+++ b/test/basics.jl
@@ -298,6 +298,62 @@ facts("Basic checks") do
         @fact value(d) --> 4*3
         @fact value(a) --> 4*3
         @fact value(b) --> 2*4*3
+
+        # check all the modified binds
+        src = Signal(false)
+        dst = Signal(false)
+        bind!(dst, !, src)
+        @fact value(dst) --> not(value(src))
+        push!(src, true)
+        step()
+        @fact value(src) --> not(value(dst))
+        push!(dst, true)
+        step()
+        @fact value(dst) --> value(src)
+
+        src = Signal(true)
+        dst = Signal(true)
+        bind!(dst, !, src, !)
+        @fact value(dst) --> not(value(src))
+        push!(src, true)
+        step()
+        @fact value(src) --> not(value(dst))
+        push!(dst, true)
+        step()
+        @fact value(dst) --> not(value(src))
+
+        src = Signal(true)
+        dst = Signal(true)
+        bind!(dst, !, src, initial=false)
+        @fact value(dst) --> value(src)
+        push!(src, true)
+        step()
+        @fact value(src) --> not(value(dst))
+        push!(dst, true)
+        step()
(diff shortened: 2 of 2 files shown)


JobJob:
Looks good at first glance, will have a closer look in the next few days.

Should `dest2src` default to `identity`?

yakir12:
That sounds like a good idea, but if `dest2src` is `identity` and we won't do the `if` closure, then why bother executing the stuff after [the `if` statement](https://github.com/yakir12/Reactive.jl/blob/68c80b4b101339ee0e60a9a6a8dcc3a762c9a85c/src/operators.jl#L386)? 
It's either execute those two lines only if `dest2src` was specified by the user, OR always execute those two lines (and the user might change `dest2src` from `identity`).
Having said that, I agree that it is by far more elegant to have `dest2src` set to `identity`...
Let me know what you prefer and I'll just change it.

JobJob:
Do we need a separate twoway flag?

yakir12:
No, cause if you want a twoway update then you *have to* specify the modifier for destination -> source.

yakir12:
Hi @JobJob,
I'm about to deploy a package which functionality depends on this PR (with the modified `bind!`). I don't mean to rush you, but this PR doesn't seem to break anything, has tests, and docs. What do you say? 
If renaming it to `bind_modified` or some such would help you decide, let me know, I'll change whatever you want.
"Worst" case scenario, I can always have this functionality in-house in my package. 
Thanks!

src/operators.jl
```diff
@@ -370,6 +370,26 @@ function bind!(dest::Signal, src::Signal, twoway=true; initial=true)
 end
 
 """
+    `bind!(dest, src2dest, src, dest2src=nothing; initial=true)`
+
+for every update to `src` also update `dest` with a modified value (using the
+function `src2dest`) and, if `dest2src` is supplied, a two-way update will hold.
+If `initial` is false, `dest` will only be updated to `src`'s modified value
+when `src` next updates, otherwise (if `initial` is true) both `dest` and `src`
+will take the respective's modiefied values immediately. Note that currently
```
JobJob: modiefied -> modified

src/operators.jl
```diff
@@ -370,6 +370,26 @@ function bind!(dest::Signal, src::Signal, twoway=true; initial=true)
 end
 
 """
+    `bind!(dest, src2dest, src, dest2src=nothing; initial=true)`
```
JobJob: maybe add the types?

src/operators.jl
```diff
@@ -370,6 +370,26 @@ function bind!(dest::Signal, src::Signal, twoway=true; initial=true)
...
+will take the respective's modiefied values immediately. Note that currently 
+this can not be `unbind!`. Also note that for a two-way bind, a false initial
+is impossible.
+"""
+function bind!(dest::Signal, src2dest::Function, src::Signal, dest2src = nothing; initial = true)
+    dest2 = map(src2dest, src)
+    bind!(dest, dest2, false, initial = initial)
+    if dest2src ≠ nothing
+        src2 = map(dest2src, dest)
+        bind!(src2, src, true, initial = initial)
```
JobJob: why is `twoway` true here?

JobJob: Won't that just mean updates to `src` modify the hidden `src2`?

src/operators.jl
```diff
@@ -370,6 +370,26 @@ function bind!(dest::Signal, src::Signal, twoway=true; initial=true)
...
 """
+    `bind!(dest, src2dest, src, dest2src=nothing; initial=true)`
+
+for every update to `src` also update `dest` with a modified value (using the
+function `src2dest`) and, if `dest2src` is supplied, a two-way update will hold.
+If `initial` is false, `dest` will only be updated to `src`'s modified value
+when `src` next updates, otherwise (if `initial` is true) both `dest` and `src`
+will take the respective's modiefied values immediately. Note that currently 
+this can not be `unbind!`. Also note that for a two-way bind, a false initial
+is impossible.
```
JobJob: > Also note that for a two-way bind, a false initial is impossible.

Is that true? Doesn't look like it from the code, but maybe I'm missing something?

src/operators.jl
JobJob: Ahh wait aren't `src2` and `src` the wrong way around here?

JobJob: Ahh I see, you have to have it in the reverse order with two-way to avoid an infinite cycle of updates (standard `bind!` handles that case)

src/operators.jl
JobJob: There's a bug in twoway bind with `initial=false`, [this line](https://github.com/JuliaGizmos/Reactive.jl/blob/master/src/operators.jl#L375) should be `bind!(src, dest, false; initial = initial)`, anyway, I'll add it after I merge this, and remove this line from the docstring.

PR #159, opened 2017-09-05
Pull request/a9c079d4

yakir12:
fixes #158 
This fixes everything. What did the trick is to just copy-paste the original `bind!` code with three slight modifications:
1.`twoway = dest2src ≠ identity`
2. in the `bind_updater` inner function, modify the `src.value` using the `src2dest` function
3. fix the glitch you found in [this line](https://github.com/JuliaGizmos/Reactive.jl/blob/master/src/operators.jl#L375) (which you still need to do: I didn't want to overstep your hospitality)

All the weirdness with initial and twoway is gone, and `unbind!` works out-of-the-box.

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -16,6 +16,7 @@ export map,
        droprepeats,
        flatten,
        bind!,
+       bindmap!,
        unbind!,
        bound_srcs,
        bound_dests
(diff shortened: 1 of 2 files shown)


yakir12:
I forgot to mention, I figured since this is like `bind!` but different, people might look for it by typing `bind` in the REPL and then tab, so it made sense to name it bind...something. So it ended up being `bindmap!`. Feel free to change it though!

PR #160, opened 2017-09-05
Pull request/97b93347

yakir12:
forgot to remove a small `unbindmap!`. Fixed that.

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -16,6 +16,7 @@ export map,
        droprepeats,
        flatten,
        bind!,
+       bindmap!,
        unbind!,
        bound_srcs,
        bound_dests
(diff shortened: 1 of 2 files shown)


JobJob:
Yeah this is probably better, what do you think about the default value for dest2src being `not_quite_identity` which you'd define as `not_quite_identity(x) = x` so you can pass `identity` if you want a non-mapped reverse.

I'm a bit concerned about the code duplication, so I think `bind!(dest, src, twoway)` should just be a call to `bindmap!(dest, identity, src, dest2src)` with `dest2src = twoway ? identity : not_quite_identity`

Will that work?

Also the method signature in the docstring is not in sync with the code, and I think putting the types there would make the docstring clearer. In the docstring probably put `dest2src=nothing`.

JobJob:
Or maybe just `dest2src = nothing` by default makes more sense.

PR #161, opened 2017-09-05
Pull request/abba71cc

yakir12:
Yea, I forgot to fix the docstring.. 
I figured that `dest2src = nothing` is simpler and makes more sense so no `like_identity` is needed. The default behavior of `bind!` for `twoway` results in `dest2src = identity` anyhow so in practice that's what happens except when the user wants to modify the binded map in a specific way.

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -16,6 +16,7 @@ export map,
        droprepeats,
        flatten,
        bind!,
+       bindmap!,
        unbind!,
        bound_srcs,
        bound_dests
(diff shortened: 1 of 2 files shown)


test/basics.jl
```diff
@@ -298,6 +298,90 @@ facts("Basic checks") do
...
+        @fact value(dst) --> 5
+        push!(dst,1)
+        step()
+        @fact value(src) --> -1
+
+        # This should work but I think line 444 in operators.jl shouldn't exit if the twoway unbind! was applied after a oneway unbind!.
+        # unbind!(dst,src) # test other way as well
+        # push!(dst,2)
+        # step()
+        # @fact value(src) --> -1
```
JobJob: Should this work now?

src/operators.jl
```diff
@@ -353,7 +370,7 @@ function bind!(dest::Signal, src::Signal, twoway=true; initial=true)
             # `true` below is for dont_remove_dead nodes - messes with active_nodes
             # TODO - check that - not sure it actually does, this may be a relic
             # of an earlier implementation which used the node's id's
-            run_push(dest, src.value, onerror_rethrow, true)
+            run_push(dest, src2dest(src.value), onerror_rethrow, true) # here is the only place where we implement the modifying function!
```
JobJob: nice 👍

test/basics.jl
```diff
@@ -298,6 +298,90 @@ facts("Basic checks") do
...
+        push!(src, 2)
+        step()
+        @fact value(dst) --> 4
+        push!(dst, 5)
+        step()
+        @fact value(src) --> 2
+        unbind!(dst,src)
+        push!(src,1)
+        step()
+        @fact value(dst) --> 5
```
JobJob: 👍

JobJob:
This is nice in the end - good work!

yakir12:
Sweet! Very happy to help. All my best work comes from copy-pasting other's work...

test/basics.jl
yakir12: I don't think so. [This line](https://github.com/JuliaGizmos/Reactive.jl/blob/5d1a8da7c6053330f2db31df6febdea028dfbe9c/src/operators.jl#L403), I think, only checks src=>dest but not dest=>src? so it just exists even if the other way around existed...

JobJob:
😄

Issue #163, opened 2017-10-27
Ajdust the frequency of every using the value of an external signal?


saolof:
Title says it all.  Would it be possible to change the time between the next triggers of every by adjusting a signal? Right now as I understand it, you can only set the timer once.

JobJob:
> It would also be nice to have a clock signal that starts/stops efficiently depending on the value of another boolean signal

See `fpswhen(switch, rate)`

> Would it be possible to change the time between the next triggers of every by adjusting a signal?

I think you might be able to implement this with these minor changes:
1) modify [fpswhen_connect](https://github.com/JuliaGizmos/Reactive.jl/blob/master/src/time.jl#L128) to take a signal rather than a constant
2) modify the existing fpswhen to `fpswhen(switch::Signal, rate::Signal)`, and add `fpswhen(switch::Signal, rate) = fpswhen(switch, Signal(rate))`
3) change the `dt = 1.0/rate` to `dt = 1.0/rate.value` and move it to be at the top `fpswhen_runner()`

PR welcome if you get it working.

saolof:
I got it working!

`function fpswhen_rate(switch, rate; name=auto_name!("fpswhen_rate", rate, switch))
    # Creates a node and sets up a timer that pushes to the node every 1.0/rate
    # seconds.
    n = Signal(Float64, 0.0, (switch,rate ,); name=name)
    fpswhen_rate_connect(rate, switch, n, name)
    n
end

function fpswhen_rate_connect(rate, switch, output, name)
    prev_time = time()
    outputref = WeakRef(output)
    switchref = WeakRef(switch)
    timer = Timer(identity, 0) # dummy timer to initialise
    function fpswhen_runner()
        # this function will run if switch gets a new value (i.e. is "active")
        # and if output is pushed to (assumed to be by the timer)
        if switch.value
            start_time = time()
            timer = setup_next_tick(outputref, switchref, start_time-prev_time, 1.0/value(rate))
            prev_time = start_time
        else
            close(timer)
        end
        switch.value
    end
    # the fpswhen_aux will start and stop the timer if switch's value updates, it'll
    # also setup the next tick once the first tick is pushed to output
    fpswhen_aux = Signal(switch.value, (switch, output); name="fpswhen runner switch: $(switch.name), output: $(output.name)")
    preserve(fpswhen_aux)
    add_action!(fpswhen_runner, fpswhen_aux)

    fpswhen_runner() # init
[...]

PR #164, opened 2017-10-28
Add function to time.jl

saolof:
Adds a new function fpswhen_rate where the rate at which the output signal is emitted is specified by a signal as opposed to a constant.

### DIFF
--- a/src/time.jl
+++ b/src/time.jl
@@ -154,6 +154,51 @@ function fpswhen_connect(rate, switch, output, name)
     finalizer(output, _->close(timer))
 end
 
+
+"""
+    fpswhen_rate(switch, rate)
+
+Same as same as fpswhen, but where the rate is a signal instead of a constant.
+"""
+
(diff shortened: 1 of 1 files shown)


JobJob:
Thanks for the PR, glad you got it working. 

I'm a bit concerned about the code duplication with this though, i.e. fpswhen_rate, and fpswhen_rate_connect are almost identical to fpswhen, and fpswhen_connect, they just take arguments with different types. 

Was there a problem with just adding another method to fpswhen with `rate` having type `Signal`, like what I wrote in https://github.com/JuliaGizmos/Reactive.jl/issues/163#issuecomment-340132351 ?

saolof:
Ah right, it should probably work as just an overload of fpswhen for signal inputs.

JobJob:
Yep, good point, and we have `value(x) = x` since https://github.com/JuliaGizmos/Reactive.jl/pull/150

PR #165, opened 2017-11-09
changes for static compilation

SimonDanisch:
mostly turned globals into constants with a concrete type.

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -10,6 +10,25 @@ const debug_memory = false # Set this to true to debug gc of nodes
 const nodeset = WeakKeyDict()
 const io_lock = ReentrantLock()
 
+const runner_task = Ref{Task}()
+
+const run_remove_dead_nodes = Ref(false)
+
+const CHANNEL_SIZE = 1024
+
+
+#run in asynchronous mode by default
+const async_mode = Ref(true)
+
+const debug_mode = Ref(false)
+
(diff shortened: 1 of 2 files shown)


src/core.jl
```diff
@@ -52,32 +71,31 @@ else
...
 
-const edges = Vector{Int}[] #parents to children, useful for plotting graphs
 
-const node_count = DefaultDict{String,Int}(0) #counts of different signals for naming
 function auto_name!(name, parents...)
     parents_str = join(map(s->s.name, parents), ", ")
     isempty(parents_str) || (name *= "($parents_str)")
-    node_count[name] += 1
-    countstr = node_count[name] > 1 ? "-$(node_count[name])" : ""
+    count = get!(node_count, name, 0) + 1
```
JobJob: `node_count` is a DefaultDict so don't need the `get!` But you could just change it to a regular Dict, and leave the `get!`, if you prefer that style.

SimonDanisch: That's what I did ;)

JobJob: Ha lol, ok

SimonDanisch: To be honest, I'm not a hundred percent sure if that was a problem. But globals turned out pretty fragile, so I tried to use only Base types for globals. And the stack traces pointed to auto_name! and this was, with other changes, the first one working.

JobJob: Sweet

JobJob:
So there's nothing here besides formatting, that node_count change, and runner_task changed to a ref? If so, looks good to me.

PR #168, opened 2017-11-16
Fix runner_task refs and Update Travis to test julia 0.6

JobJob:
Also fix the `every` test that always fails

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ os:
     - linux
     - osx
 julia:
-    - 0.5
+    - 0.6
     - nightly
 sudo: false
 notifications:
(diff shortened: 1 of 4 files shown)


JobJob:
@SimonDanisch some of the test failures in your last PR were actually legit and due to the change of `Reactive.runner_task` to a Ref, this fixes that.

It also fixes the issue with the `every()` test failing on julia 0.7 - that was caused by [a recent change](https://github.com/JuliaLang/julia/commit/bbf558458583bce45256effcb78577256cf92f58) to `isapprox` - which IMO was broken when used with just `atol` before.

This PR only changes tests/travis. All tests are passing again now on 0.6, 0.7 🥂

PR #170, opened 2017-11-23
Add dispatch to fpswhen for time-varying rate

saolof:
Cleaned up version of PR 164. Adds a dispatch to fpswhen so that the rate is allowed to be a signal, instead of defining a separate function. Makes a slight change to fpswhen_connect in order to make it cover both cases.

### DIFF
--- a/src/time.jl
+++ b/src/time.jl
@@ -117,6 +117,13 @@ function fpswhen(switch, rate; name=auto_name!("$rate fpswhen", switch))
     n
 end
 
+function fpswhen(switch, rate::Signal; name=auto_name!("fpswhen_rate", rate, switch))
+    #Dispatch for the case where the rate is a time-varying signal.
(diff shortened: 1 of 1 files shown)


src/time.jl
```diff
@@ -117,6 +117,13 @@ function fpswhen(switch, rate; name=auto_name!("$rate fpswhen", switch))
     n
 end
 
+function fpswhen(switch, rate::Signal; name=auto_name!("fpswhen_rate", rate, switch))
+    #Dispatch for the case where the rate is a time-varying signal.
+    n = Signal(Float64, 0.0, (switch,rate ,); name=name)
```
JobJob: Having `rate` as a parent of `n` seems reasonable. I'm curious, was it needed, or you just thought it was sensible?

saolof: The latter. I haven't tried it without rate as a parent though.

PR #171, opened 2017-11-23
Determine CHANNEL_SIZE from environment

### LINKED ISSUE #169: Channel size
I have a somewhat more complicated Makie plot and run into channel size issues. I see that the channel size is hardcoded to 1024. Should it be possible to increase that?

Edit: I guess the outcommented lines strain the message queue are the loops are looping over 500 line vectors. https://github.com/mschauer/Bridge.jl/blob/partial2/example/makie.jl#L23

mschauer:
Fixes #169 in the way proposed by @JobJob .

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -14,7 +14,7 @@ const runner_task = Ref{Task}()
 
 const run_remove_dead_nodes = Ref(false)
 
-const CHANNEL_SIZE = 1024
+const CHANNEL_SIZE = parse(Int, get(ENV, "REACTIVE_CHANNEL_SIZE", "1024"))
 
 
 #run in asynchronous mode by default
@@ -256,7 +256,8 @@ end
 function async_push!(n, x, onerror=print_error)
     taken = Base.n_avail(_messages)
     if taken >= CHANNEL_SIZE
-        warn("Message queue is full. Ordering may be incorrect.")
+        warn("Message queue is full. Ordering may be incorrect. "*
+        "Channel size can be increased by setting `ENV[\"REACTIVE_CHANNEL_SIZE\"] = ...`")
         @async put!(_messages, Message(n, x, onerror))
     else
         put!(_messages, Message(n, x, onerror))
--- a/src/core.jl
+++ b/src/core.jl
@@ -14,8 +14,7 @@ const runner_task = Ref{Task}()
 
 const run_remove_dead_nodes = Ref(false)
 
-const CHANNEL_SIZE = parse(Int, get(ENV, "REACTIVE_CHANNEL_SIZE", "1024"))
-
+const CHANNEL_SIZE = Ref(1024)
 
 #run in asynchronous mode by default
 const async_mode = Ref(true)
@@ -224,7 +223,7 @@ immutable Message
 end
 
 # Global channel for signal updates
-const _messages = Channel{Nullable{Message}}(CHANNEL_SIZE)
+const _messages = Channel{Nullable{Message}}(CHANNEL_SIZE[])
 
 run_async(async::Bool) = (async_mode[] = async)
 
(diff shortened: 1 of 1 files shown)


JobJob:
This is really good, but you're not based off master! Can you fix it up? prob easiest just to manually:
```
git format-patch -1 4eda8253161600e053fc4dbe3970980bf503b9b0 #sha of your commit
git checkout master
git pull
git branch channelsize2
patch -p1 < 0001-Determine-CHANNEL_SIZE-from-environment.patch
```

mschauer:
Ah, I started off with `git checkout master` without a fetch. Fixed here.

src/core.jl
```diff
@@ -256,7 +256,8 @@ end
 function async_push!(n, x, onerror=print_error)
     taken = Base.n_avail(_messages)
     if taken >= CHANNEL_SIZE
-        warn("Message queue is full. Ordering may be incorrect.")
+        warn("Message queue is full. Ordering may be incorrect. "*
+        "Channel size can be increased by setting `ENV[\"REACTIVE_CHANNEL_SIZE\"] = ...`")
```
JobJob: Nice touch 😄 
maybe we should also add "by setting `ENV[\"REACTIVE_CHANNEL_SIZE\"] = ...`" before `using Reactive`" ?

src/core.jl
```diff
@@ -14,7 +14,7 @@ const runner_task = Ref{Task}()
 
 const run_remove_dead_nodes = Ref(false)
 
-const CHANNEL_SIZE = 1024
+const CHANNEL_SIZE = parse(Int, get(ENV, "REACTIVE_CHANNEL_SIZE", "1024"))
```
JobJob: Nice, didn't realise ENV converts to strings auto-magically.

src/core.jl
mschauer: Hm..., that won't be enough because of pre-compilation freezing the value into the constant.

JobJob: oh shit, you're right... maybe we need to change `CHANNEL_SIZE` to be a Ref ?

JobJob: As in
```
const CHANNEL_SIZE = Ref(1024)
CHANNEL_SIZE[] = parse(Int, get(ENV, "REACTIVE_CHANNEL_SIZE", "1024"))
```
or something

JobJob: maybe `haskey(ENV, "REACTIVE_CHANNEL_SIZE") && (CHANNEL_SIZE[] = parse(Int, ENV["REACTIVE_CHANNEL_SIZE"])` is better (less repetition of 1024)

JobJob: Oh, that needs to go in an `__init__()`

JobJob: https://github.com/JuliaGizmos/Reactive.jl/blob/master/src/core.jl#L417

mschauer: What is the advantage of
```
const CHANNEL_SIZE = Ref(parse(Int, get(ENV, "REACTIVE_CHANNEL_SIZE", "1024")))
```
 over 
```
CHANNEL_SIZE = parse(Int, get(ENV, "REACTIVE_CHANNEL_SIZE", "1024"))
```
?

JobJob: Because then we can set the Ref in the `__init__()` function. And it will be updated based on the current value of `ENV["REACTIVE_CHANNEL_SIZE"]` on `using Reactive`

JobJob: I just tried out `const CHANNEL_SIZE = Ref(1024)`
and in `__init__()`

```
    haskey(ENV, "REACTIVE_CHANNEL_SIZE") && 
        (CHANNEL_SIZE[] = parse(Int, ENV["REACTIVE_CHANNEL_SIZE"]))
    runner_task[] = @async run()
```
plus changing lines ~226, ~257 from `CHANNEL_SIZE` to `CHANNEL_SIZE[]` and it seems to respect my `ENV["REACTIVE_CHANNEL_SIZE"]`

JobJob: Oh, I missed that your CHANNEL_SIZE is not `const`. The const is an advantage for static compilation, and because non-const globals can be slow.

mschauer: Ah, I get it, `const x = Ref` makes sure that can infer `Int` because only the value but not the type can change. https://github.com/JuliaLang/julia/issues/13817#issuecomment-152157853

JobJob:
Awesome. Thanks very much!

PR #175, opened 2018-03-15
Fix droprepeat example

Paethon:
Fixes an error in the example for droprepeats (I think!?)


Paethon:
I only changed the documentation, so I am pretty sure the failing tests have nothing to do with it.

shashi:
thanks!

PR #178, opened 2018-06-01
changes for 0.7

SimonDanisch:


SimonDanisch:
Hope this passes tests on 0.6 as well... I do a bit shady things to keep compat to 0.6 - don't have the patience to read into Compat.jl or how to replace Nullable correctly :P

SimonDanisch:
* gets rid of FactCheck
* drop 0.5 support
* remove Nullable and use MaybeMessage instead (lol)
* use fetch instead of wait on 0.7, continue using wait on 0.6
* fixes are for  v"0.7.0-alpha.2" only

SimonDanisch:
Hmm.... what's up with linux 0.7!? Can someone reproduce?

SimonDanisch:
Ok, might this be a compile time regression messing up our timing? And that's why I can't reproduce, because I'm on a pretty fast machine? Anyone wants to try this out locally?

SimonDanisch:
Should I just go ahead and merge? The only failure is 0.7 on linux... so the situation is much better then what we currently have, and I'm 90% sure it's a weird timing issue!
@shashi @JobJob

JobJob:
I vote yes\*
\*: haven't looked at this at all\*\*
\*\*: I believe in you @SimonDanisch 😄

timholy:
One problem: the REQUIRE file wasn't updated to `julia 0.7.0-alpha`. That means I am getting a Travis run like this one: https://travis-ci.org/JuliaGizmos/GtkReactive.jl/jobs/392323531, in which you can see that it downloaded Reactive v0.7.0 (which claims to drop support for julia 0.6) despite the fact that it was running julia 0.6.

@SimonDanisch, can you fix? I guess I'd propose that the best strategy might be to release a 0.6.1 including the commits up to Nov 2017 and a manual PR against METADATA to fix the requirements for 0.7.

timholy:
I went ahead and did the 0.6.1 release, just need to fix the `requires` in METADATA for the 0.7 release and then update the `REQUIRE` file here.

timholy:
Oh wait, while the release's title includes "drop support for 0.6," reading the above more carefully it seems like it still works on 0.6? I won't take any further steps without input from Simon (sorry for any confusion).

PR #183, opened 2018-08-22
fix remaining depwarns - actually drop 0.6

SimonDanisch:


SimonDanisch:
The failure seems to be the usual isapprox failure with too little wiggle room for osx

PR #185, opened 2018-08-30
fix show method for Signal

piever:
Otherwise `show` would throw on a `Signal` with value `nothing`.


SimonDanisch:
should we use repr here?

piever:
Not sure:

```julia
julia> s = "test";

julia> repr(s)
"\"test\""

julia> s == repr(s)
false
```

With `repr` we'd show all strings with the `'\"'` character.

